PR #8, opened 2026-04-15
Add RALPH-CC.md: run ralph loop inside Claude Code

### LINKED ISSUE #7: Support running the ralph loop natively inside a Claude Code session
## Problem

`ralph.sh` depends on `claude -p` (headless mode) to spawn a fresh Claude per iteration. Claude Code may remove `-p` or restrict it to API-only auth (no OAuth). If that happens, `ralph.sh` breaks.

## Proposal

Provide a drop-in alternative to `./ralph.sh --max-iterations N` that runs entirely inside a live Claude Code chat session, using the `Agent()` subagent tool instead of `claude -p`.

Constraints:
- Per-iteration semantics must stay identical to `ralph.sh`.
- Same `PROMPT.md`, same `tasks.json`, same `.lnb/` notebook, same `archive/` behavior.
- No changes to `PROMPT.md`.
- No new features beyond what `ralph.sh` already has.

## Shape

- Factor `ralph.sh`'s helper functions (`read_task_meta`, `archive_previous_run`, `ensure_notebook`, `log_to_notebook`, `query_recent_history`, `build_prompt`) into a shared `ralph-lib.sh` so both runners stay in sync.
- New `ralph-prep.sh`: per-iteration bookkeeping that prints the filled prompt to stdout. Intended to be called from a Claude Code session via `Bash`, then piped to `Agent()`.
- New `RALPH-CC.md`: driver doc the main session follows — the loop procedure, stop conditions, and permission-mode notes.

carbonscott:
## Summary

- Factor `ralph.sh`'s helper functions into a shared `ralph-lib.sh` so future runners can reuse the same bookkeeping (archive, notebook init, prompt fill, history query).
- Add `RALPH-CC.md` + `ralph-prep.sh`: a drop-in alternative to `./ralph.sh --max-iterations N` that runs entirely inside a live Claude Code chat session, using `Agent()` instead of headless `claude -p`. Per-iteration semantics, `PROMPT.md`, `tasks.json`, and `.lnb/` are identical to the existing flow.
- Add `tests/` sandbox scaffold for smoke-testing both modes.

## Motivation

Closes #7. `claude -p` (headless mode) may be removed or restricted to API-only auth. `ralph.sh` depends on it. This PR provides an equivalent that runs natively inside Claude Code so the workflow survives.

Key design decisions:

- **`PROMPT.md` is untouched.** The existing SIGNAL step (`<promise>DONE</promise>` / `<promise>ALL_DONE</promise>`) works as-is when the subagent's final message gets returned via the `Agent` tool.
- **No orchestrator subagent layer.** The main Claude Code session *is* the orchestrator — it runs the loop, spawns workers, checks promises. The ~50 tokens of subagent narration that leaks into the main session per iteration is acceptable (~1k tokens over 20 iterations).
- **Shared-lib refactor.** `ralph.sh` and `ralph-prep.sh` both source `ralph-lib.sh`, so the two runners stay in sync forever.
- **Stop-on-no-marker.** If a subagent returns without any promise marker, the main session stops and logs a blocker. Small deviation from `ralph.sh` (which only stops on non-zero exit), but the subagent model has no exit code.

## Behavior differences from `ralph.sh`

| Aspect | `ralph.sh` | RALPH-CC.md flow |
|---|---|---|
| Entry point | `./ralph.sh --max-iterations N` | Chat: "follow RALPH-CC.md, max-iterations N" |
| Per-iter agent | `claude -p --permission-mode acceptEdits` | `Agent(subagent_type="general-purpose")` |
| Permission mode | CLI flag | Inherited from main session |
| Intermediate display | stream-json via `format_stream` | Claude Code's native subagent UI |
| Stop on no marker | no | yes |
| Main-session context cost | 0 | ~50 tokens/iter |
| State storage (`tasks.json`, `.lnb/`, `archive/`) | same | same |

## Commits

1. `refactor(ralph): extract helpers into ralph-lib.sh` — pure refactor, -86/+4 in `ralph.sh`, new `ralph-lib.sh`. Also routes diagnostic output in `archive_previous_run` and `ensure_notebook` to stderr so stdout-capturing callers (like `ralph-prep.sh`) stay clean. No behavior change for `ralph.sh`.
2. `feat: add RALPH-CC.md flow for running ralph inside Claude Code` — adds `ralph-prep.sh`, `RALPH-CC.md`, `.gitignore`.
3. `test: add tests/ sandbox scaffold for smoke-testing ralph` — adds `tests/README.md`, `tests/setup-sandbox.sh`, `tests/tasks.json`.

## Test plan

- [x] **`ralph.sh` regression:** `./tests/setup-sandbox.sh && cd <sandbox> && ./ralph.sh --max-iterations 3` — should complete both fixture stories and report `ALL_DONE` within 2 iterations, same as before.
- [x] **New RALPH-CC.md flow:** `./tests/setup-sandbox.sh`, then `cd <sandbox> && claude` in `acceptEdits` mode, then in the session: `follow RALPH-CC.md, max-iterations 3`. Expect `hello.txt` created with both lines, `tasks.json` stories flipped to `passes: true`, `.lnb/` populated, `ALL_DONE` reported.
- [x] **Stdout cleanliness:** `./ralph-prep.sh --iteration 1 > /tmp/prompt.out` (in a sandbox with a fixture `tasks.json`) — `/tmp/prompt.out` should contain only the filled prompt, no `Initialized lab notebook` or other diagnostic noise.
- [ ] **Archive-on-branch-change:** run once, change the `branch` field in `tasks.json`, run again — previous run should be archived under `archive/<date>-<old-branch>/`.

## Out of scope (follow-ups)

- `README.md` update advertising the new mode.
- Additional `.gitignore` entries (`.ralph-last-branch`, `archive/`).

### DIFF
--- a/RALPH-CC.md
+++ b/RALPH-CC.md
@@ -46,7 +46,16 @@ minimum specified.
 
 1. Parse `max-iterations` and `task-file` from the user's message.
    Apply defaults for anything unspecified.
-2. Announce once: `"Starting ralph loop, max-iterations=N, task-file=X"`.
+2. Derive the notebook context once, so every subsequent log call uses
+   the same value:
+
+   ```
+   Bash("jq -r '.branch // .project // \"ralph-dev\"' <task-file>")
+   ```
+
+   Capture stdout into `<context>`. This mirrors how `ralph-prep.sh`
+   derives it, so both runners agree on the context slug.
+3. Announce once: `"Starting ralph loop, max-iterations=N, task-file=X, context=<context>"`.
    One line, nothing more.
 
 ### Loop
@@ -56,11 +65,14 @@ For `i` in `1..max-iterations`:
 1. **Build the prompt.** Call:
 
    ```
-   Bash("./ralph-prep.sh --iteration i --task-file <task-file>")
+   Bash("./ralph-prep.sh --iteration i --max-iterations N --task-file <task-file>")
    ```
 
-   Capture the tool result's stdout. This is the filled prompt. Do not
-   read, quote, or summarize it — just hold it for the next step.
+   Pass the same `N` you parsed in Setup so the start log entry records
+   the cap alongside the iteration number (matches `ralph.sh`'s log
+   format). Capture the tool result's stdout. This is the filled
+   prompt. Do not read, quote, or summarize it — just hold it for the
+   next step.
 
 2. **Spawn the worker.** Call:
 
@@ -76,12 +88,15 @@ For `i` in `1..max-iterations`:
    ending with either `<promise>DONE</promise>` or
    `<promise>ALL_DONE</promise>`.
 
-3. **Check the return.** Inspect the subagent's final message:
+3. **Check the return.** Inspect the subagent's final message. **Check
+   `ALL_DONE` first — `<promise>ALL_DONE</promise>` contains
+   `<promise>DONE</promise>` as a substring, so the order is
+   load-bearing.** (Same reason `ralph.sh`'s grep checks `ALL_DONE`
+   before `DONE` at ralph.sh:~159.)
 
    - Contains `<promise>ALL_DONE</promise>`:
-     - Call `Bash("LAB_NOTEBOOK_DIR=<notebook> lab-notebook emit --context <context> --type done --tags ralph-harness 'ralph-cc: all stories complete at iteration i'")`.
-       (Context comes from the task file's `branch` field; if unsure,
-       omit the emit — not critical.)
+     - Call `Bash("LAB_NOTEBOOK_DIR=<notebook> lab-notebook emit --context <context> --type done --tags ralph-harness 'ralph-cc: all stories complete at iteration i'")`
+       using the `<context>` derived in Setup step 2.
      - Break the loop. Go to "Post-loop".
    - Contains `<promise>DONE</promise>`:
      - Say exactly: `"iteration i: DONE, continuing"`. One line.
@@ -94,7 +109,7 @@ For `i` in `1..max-iterations`:
 ### Post-loop
 
 1. If the loop ended on `ALL_DONE`, optionally query the notebook for a
-   summary:
+   summary, using the `<context>` derived in Setup step 2:
 
    ```
    Bash("LAB_NOTEBOOK_DIR=<notebook> lab-notebook sql \"SELECT ts, type, issue, substr(content,1,200) FROM entries WHERE context='<context>' AND type IN ('start','done','impl','blocker') ORDER BY ts\"")
--- a/RALPH-CC.md
+++ b/RALPH-CC.md
@@ -1,0 +1,145 @@
+# Ralph Loop — Claude Code Session Edition
+
+Drop-in alternative to `./ralph.sh --max-iterations N`, designed to run
+entirely inside a live Claude Code chat session using the `Agent()`
+subagent tool instead of headless `claude -p`. Use this when `-p` mode
+is unavailable or restricted.
+
+Same `PROMPT.md`, same `tasks.json`, same `.lnb/` notebook, same
+`archive/` directory, same per-iteration semantics. The only thing that
+changes is who spawns the per-iteration agent.
+
+## How to invoke
+
+In a Claude Code chat, type one of:
+
+> follow RALPH-CC.md, max-iterations 10
+
+> follow RALPH-CC.md, max-iterations 5, task-file tasks.json
+
+Defaults when unspecified:
+
+- `max-iterations` = 10
+- `task-file` = `tasks.json`
+- `prompt` = `PROMPT.md`
+- `notebook` = `.lnb`
+
+## Prerequisites
+
+1. The main Claude Code session must be in `acceptEdits` permission mode
+   (or you must approve each subagent tool call manually). Subagents
+   inherit the parent session's permission mode. This replaces
+   `ralph.sh`'s `--permission-mode acceptEdits` flag.
+2. `ralph-prep.sh`, `ralph-lib.sh`, and `PROMPT.md` must exist in the
+   current directory. `tasks.json` must exist and contain the stories to
+   work through.
+3. `ralph-prep.sh` must be executable (`chmod +x ralph-prep.sh`).
+4. `jq` and `lab-notebook` must be on `$PATH` (same as for `ralph.sh`).
+
+## Procedure for the main Claude Code session
+
+When the user invokes this file, the main session (you) must do the
+following. Do not deviate, do not add steps, do not narrate beyond the
+minimum specified.
+
+### Setup
+
+1. Parse `max-iterations` and `task-file` from the user's message.
+   Apply defaults for anything unspecified.
+2. Announce once: `"Starting ralph loop, max-iterations=N, task-file=X"`.
+   One line, nothing more.
+
+### Loop
+
+For `i` in `1..max-iterations`:
+
+1. **Build the prompt.** Call:
+
+   ```
+   Bash("./ralph-prep.sh --iteration i --task-file <task-file>")
+   ```
+
+   Capture the tool result's stdout. This is the filled prompt. Do not
+   read, quote, or summarize it — just hold it for the next step.
+
+2. **Spawn the worker.** Call:
+
+   ```
+   Agent(
+       subagent_type="general-purpose",
+       description="ralph iteration i",
+       prompt=<stdout from step 1>
+   )
+   ```
+
+   The subagent will complete one story and return a final message
+   ending with either `<promise>DONE</promise>` or
+   `<promise>ALL_DONE</promise>`.
+
+3. **Check the return.** Inspect the subagent's final message:
+
+   - Contains `<promise>ALL_DONE</promise>`:
+     - Call `Bash("LAB_NOTEBOOK_DIR=<notebook> lab-notebook emit --context <context> --type done --tags ralph-harness 'ralph-cc: all stories complete at iteration i'")`.
+       (Context comes from the task file's `branch` field; if unsure,
+       omit the emit — not critical.)
+     - Break the loop. Go to "Post-loop".
+   - Contains `<promise>DONE</promise>`:
+     - Say exactly: `"iteration i: DONE, continuing"`. One line.
+     - Continue to next iteration.
+   - Neither marker:
+     - Call `Bash(...)` to log a blocker entry (same shape as above but
+       `--type blocker` and `"ralph-cc: iteration i ended without promise"`).
+     - Break the loop. Go to "Post-loop".
+
+### Post-loop
+
+1. If the loop ended on `ALL_DONE`, optionally query the notebook for a
+   summary:
+
+   ```
+   Bash("LAB_NOTEBOOK_DIR=<notebook> lab-notebook sql \"SELECT ts, type, issue, substr(content,1,200) FROM entries WHERE context='<context>' AND type IN ('start','done','impl','blocker') ORDER BY ts\"")
+   ```
+
+   Report a 3-5 sentence summary to the user covering what each story
+   accomplished.
+
+2. If the loop ended on `max-iterations`, report:
+   `"Hit iteration cap (N). Stopped. Check .lnb/ for progress."`
+
+3. If the loop ended on a missing marker, report:
+   `"Iteration i returned without a promise marker. Stopped. Check .lnb/ for the blocker entry and the subagent's final message."`
+
+## Stop conditions (summary)
+
+- `<promise>ALL_DONE</promise>` seen → success.
+- `max-iterations` reached → capped.
+- No promise marker in a subagent return → blocker.
+
+## Context budgeting
+
+Each iteration adds ~50 tokens (the subagent's final narration plus the
+promise marker) to the main session's context. A 20-iteration run is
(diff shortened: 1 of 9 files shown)


ralph-lib.sh
```diff
@@ -0,0 +1,90 @@
...
+    fi
+}
+
+# --- Notebook helpers ---
+ensure_notebook() {
+    if [[ ! -d "$NOTEBOOK_DIR" ]]; then
+        if [[ -f "$SCRIPT_DIR/coding-dev.yaml" ]]; then
+            mkdir -p "$NOTEBOOK_DIR"
+            cp "$SCRIPT_DIR/coding-dev.yaml" "$NOTEBOOK_DIR/schema.yaml"
+            LAB_NOTEBOOK_DIR="$NOTEBOOK_DIR" lab-notebook init "$NOTEBOOK_DIR" >&2 2>/dev/null || true
```
carbonscott: Nit: The redirect order `>&2 2>/dev/null` is non-idiomatic and easy to misread. Processed left-to-right, the net effect is: `lab-notebook init`'s stdout goes to the *caller's stderr* (not `/dev/null`), and only its stderr is silenced. That's actually what you want here (stdout must stay clean for `ralph-prep.sh`), so functionally it's fine — but a future maintainer will assume it silences both streams. Two options:

1. If you want to silence both (match the old `ralph.sh` spirit): `>/dev/null 2>&1`
2. If you really want diagnostics on the caller's stderr: keep `>&2` and drop the `2>/dev/null` so real errors aren't swallowed — or add a comment explaining the intent.

Same thing on line 46 for the `--local` branch. Low severity because the existing behavior works for both callers; just flagging so the intent is explicit.

ralph-lib.sh
```diff
@@ -0,0 +1,90 @@
...
+            --context "$CONTEXT" --type "$entry_type" \
+            --branch "${BRANCH:-}" --tags "ralph-harness" \
+            "$message" 2>/dev/null || true
+    fi
+}
+
+query_recent_history() {
+    if command -v lab-notebook &>/dev/null && [[ -d "$NOTEBOOK_DIR" ]]; then
+        LAB_NOTEBOOK_DIR="$NOTEBOOK_DIR" lab-notebook sql \
+            "SELECT ts, type, issue, substr(content,1,200) FROM entries WHERE context='$CONTEXT' ORDER BY ts DESC LIMIT 10" \
```
carbonscott: FYI: `$CONTEXT` is interpolated raw into a single-quoted SQL string. A `tasks.json` with `"branch": "foo'; DROP TABLE entries; --"` would break out. Since `tasks.json` is author-controlled this is low severity, but if you ever start accepting external task files it becomes real.

Consider parameterizing the query instead, e.g.:

```bash
LAB_NOTEBOOK_DIR="$NOTEBOOK_DIR" lab-notebook sql \
    "SELECT ts, type, issue, substr(content,1,200) FROM entries WHERE context = ? ORDER BY ts DESC LIMIT 10" \
    "$CONTEXT"
```

(if `lab-notebook sql` supports positional params — check its docs). Otherwise at least escape single quotes in `$CONTEXT` before interpolating. Not blocking this PR since the vulnerability is pre-existing (this function moved here verbatim from `ralph.sh`), but worth a follow-up.

ralph-prep.sh
```diff
@@ -0,0 +1,96 @@
...
+
+# --- Parse arguments ---
+while [[ $# -gt 0 ]]; do
+    case $1 in
+        --prompt)          PROMPT_FILE="$2"; shift 2 ;;
+        --task-file)       TASK_FILE="$2"; shift 2 ;;
+        --notebook)        NOTEBOOK_DIR="$2"; shift 2 ;;
+        --context)         CONTEXT="$2"; shift 2 ;;
+        --archive-dir)     ARCHIVE_DIR="$2"; shift 2 ;;
+        --iteration)       ITERATION="$2"; shift 2 ;;
```
carbonscott, line 54: FYI: `--iteration N` is currently only used in the `"start"` log entry on line 92. That's fine, but worth calling out in `RALPH-CC.md` so the main session knows to always pass `--iteration i` (step 1 of the loop already does this). If it's ever forgotten, the notebook's start entries will all say `iteration 1`, which would be silently misleading rather than broken. Nothing to change in this PR.

ralph-prep.sh
```diff
@@ -0,0 +1,96 @@
...
+fi
+
+# --- Bookkeeping ---
+# archive_previous_run and ensure_notebook are both idempotent — safe to
+# call on every iteration. archive_previous_run only actually archives
+# when BRANCH differs from $LAST_BRANCH_FILE's contents.
+archive_previous_run
+ensure_notebook
+
+log_to_notebook "start" "ralph-cc: starting iteration $ITERATION"
```
carbonscott: Consider: `ralph.sh` logs `"ralph.sh: starting iteration $i/$MAX_ITERATIONS"`, which includes the cap. The RALPH-CC flow loses `$MAX_ITERATIONS` because it lives in the main Claude Code session, not in `ralph-prep.sh`. If you want symmetry, you could add a `--max-iterations N` flag to `ralph-prep.sh` that's used only for this log message, or pass it via an env var. Minor — just makes post-hoc notebook queries prettier. Not blocking.

RALPH-CC.md
```diff
@@ -0,0 +1,145 @@
...
+   )
+   ```
+
+   The subagent will complete one story and return a final message
+   ending with either `<promise>DONE</promise>` or
+   `<promise>ALL_DONE</promise>`.
+
+3. **Check the return.** Inspect the subagent's final message:
+
+   - Contains `<promise>ALL_DONE</promise>`:
```
carbonscott, line 100: Nit: The stop-condition bullets are listed in the right order (ALL_DONE before DONE), which matters because `ALL_DONE` contains the substring `DONE`. It's worth one explicit sentence saying "check in this order" — a fresh session reading the doc might reorder them in its head. `ralph.sh` has an explicit comment `# Check for promises (ALL_DONE first since it also contains DONE)` on line 158 that makes this obvious; mirroring that note here would be cheap insurance.

RALPH-CC.md
```diff
@@ -0,0 +1,145 @@
...
+     - Call `Bash("LAB_NOTEBOOK_DIR=<notebook> lab-notebook emit --context <context> --type done --tags ralph-harness 'ralph-cc: all stories complete at iteration i'")`.
+       (Context comes from the task file's `branch` field; if unsure,
```
carbonscott: Nit: "Context comes from the task file's `branch` field; if unsure, omit the emit — not critical." This is the one piece of ambiguity in an otherwise tight procedure. The main session *does* need to know `<context>` for the optional post-loop summary query too (line 114). Since `ralph-prep.sh` already derives `CONTEXT` (and the main session could `jq -r '.branch // .project // "ralph-dev"' tasks.json`), consider either:

1. Explicitly tell the main session to run `context=$(jq -r '.branch // .project // "ralph-dev"' <task-file>)` as part of Setup, or
2. Have `ralph-prep.sh` emit the resolved context on stderr (e.g. `echo "context=$CONTEXT" >&2` on first iteration) so the main session can grab it.

Either would remove the "if unsure, omit the emit" fallback, which is the kind of fuzzy instruction that causes divergent behavior across sessions.

RALPH-CC.md
```diff
@@ -0,0 +1,145 @@
...
+
+1. The main Claude Code session must be in `acceptEdits` permission mode
+   (or you must approve each subagent tool call manually). Subagents
+   inherit the parent session's permission mode. This replaces
+   `ralph.sh`'s `--permission-mode acceptEdits` flag.
+2. `ralph-prep.sh`, `ralph-lib.sh`, and `PROMPT.md` must exist in the
+   current directory. `tasks.json` must exist and contain the stories to
+   work through.
+3. `ralph-prep.sh` must be executable (`chmod +x ralph-prep.sh`).
+4. `jq` and `lab-notebook` must be on `$PATH` (same as for `ralph.sh`).
```
carbonscott, line 40: Consider: Prerequisite 4 says `jq` and `lab-notebook` must be on `$PATH`. `ralph-prep.sh` already tolerates a missing `lab-notebook` (wrapped in `command -v` checks), but will hard-fail on a missing `jq` because `read_task_meta` uses it and `set -e` is on. Might be worth either adding an explicit `command -v jq` check at the top of `ralph-prep.sh` with a friendly error, or documenting the hard vs. soft requirement here. Minor.

tests/setup-sandbox.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+REPO_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+SANDBOX="${TMPDIR:-/tmp}/ralph-smoke-$(date +%Y%m%d-%H%M%S)"
+
+mkdir -p "$SANDBOX"
+
+for f in PROMPT.md ralph.sh ralph-prep.sh ralph-lib.sh RALPH-CC.md; do
+    ln -s "$REPO_DIR/$f" "$SANDBOX/$f"
```
carbonscott, line 10: Nit: Symlinks rather than copies means the sandbox tracks uncommitted edits in the source tree — the README already calls this out as "by design", which is good. One gotcha worth noting in the README: if a test run writes `.ralph-last-branch` into the sandbox, it's a real file (not a symlink), so it won't pollute the source repo — but `.lnb/` and `archive/` also land in the sandbox. Which is what we want. No action needed, just confirming the scaffold behaves as intended.

carbonscott (commented):
## Review: Add RALPH-CC.md — Claude Code session runner

Nice change. The refactor is clean, the new session-mode runner is well-scoped, and the factoring into `ralph-lib.sh` is behavior-preserving for `ralph.sh` as far as I can see. The PR description is excellent and accurately reflects the diff. Approving.

### Summary of what I checked

- **`ralph.sh` refactor is behavior-preserving.** The removed block (`read_task_meta`, `archive_previous_run`, `ensure_notebook`, `log_to_notebook`, `query_recent_history`, `build_prompt`, plus `LAST_BRANCH_FILE=".ralph-last-branch"`) moves verbatim to `ralph-lib.sh`. `LAST_BRANCH_FILE` is reassigned above the `source` line, and the helper functions are sourced before first use (line 117-118 `archive_previous_run` / `ensure_notebook`). The `read_task_meta` call site (line 88-89) correctly sits after the `source`, so function availability is fine. Nothing appears lost.

- **Variable-scoping contract in `ralph-lib.sh` is documented** (the header comment names every global the functions depend on). Both callers set them: `ralph.sh` sets all required globals before `source`, and `ralph-prep.sh` sets `SCRIPT_DIR` / `PROMPT_FILE` / `TASK_FILE` / `NOTEBOOK_DIR` / `CONTEXT` / `ARCHIVE_DIR` / `LAST_BRANCH_FILE` before sourcing, then resolves `BRANCH` / `PROJECT` after. Good.

- **Stdout/stderr discipline in `ralph-prep.sh` is correct.** `archive_previous_run`, `ensure_notebook`, `log_to_notebook`, and `query_recent_history`'s error-fallback all route to stderr or `/dev/null`, and the only thing printed to stdout is `build_prompt`'s final `echo`. Captured prompt should be clean.

- **Idempotency of per-iter bookkeeping is fine.** `archive_previous_run` is guarded by the `BRANCH != last_branch` check and updates `LAST_BRANCH_FILE` on every call, so after iteration 1 it's a no-op. `ensure_notebook` is guarded by `[[ ! -d "$NOTEBOOK_DIR" ]]`, so after iteration 1 it's a no-op. Safe to call every iteration.

- **`build_prompt` is safe against adversarial tasks.json.** The substitution uses pure bash parameter expansion (`${prompt//pattern/replacement}`), not `eval` or a subshell with expanded content, so even a task file containing `$(rm -rf /)` or backticks would be substituted literally. Good.

- **Archive-on-branch-change logic** looks correct on inspection: guards on `$TASK_FILE` + `$LAST_BRANCH_FILE` existence, only archives when `BRANCH` is non-empty and differs from the recorded previous branch, writes `$BRANCH` after the (possible) archive. FYI the PR's test-plan box for this is unchecked; a follow-up smoke test that runs prep twice with different `branch` fields would fully validate it.

- **`RALPH-CC.md` procedure** is unambiguous enough for a fresh Claude Code session to follow: defaults are stated, prerequisites are listed, the stop conditions are enumerated, and ordering (ALL_DONE check before DONE) is correct because the bullet order matches the priority. Also appreciated the explicit "~50 tokens/iter" budgeting note and the side-by-side table of differences from `ralph.sh`.

### Findings

All non-blocking — one minor correctness nit, one potential-but-low-risk quoting concern, and a couple of small suggestions. See inline comments.

### Good things worth calling out

- Pure refactor of `ralph.sh` (one commit factoring out helpers, second commit adding the new runner) keeps review tractable.
- Header comment on `ralph-lib.sh` documenting the global contract is exactly right — without it a future reader would have to reverse-engineer which callers set which variables.
- Sending `ralph-prep.sh` diagnostics to stderr so stdout stays clean for prompt capture is the right choice, and `set -euo pipefail` + the `|| true` guards in the notebook calls keep it from aborting when `lab-notebook` is unavailable.
- The smoke-test scaffold (`tests/setup-sandbox.sh` symlinking the scripts) correctly exercises the current source tree without needing a separate install step.

PR #11, opened 2026-04-15
Split into shared/ cc-headless/ cc/ and use lab-notebook --template-path

### LINKED ISSUE #10: Restructure layout: shared/ cc-headless/ cc/ and native coding-dev schema
## Problems

Three related issues in the flat repo layout surfaced together:

### 1. README bootstrap disagrees with `ensure_notebook`

The README quick-start told users to:

```bash
mkdir -p .lnb && cp coding-dev.yaml .lnb/schema.yaml
lab-notebook init .lnb
```

This was both **broken** (the old `lab-notebook init <path>` had a path-doubling quirk that ghosted the pre-placed `schema.yaml`) and **redundant** (`ensure_notebook` in `ralph-lib.sh` already did init + schema install + rebuild on first iteration). Two bootstrap paths for the same thing, one of them wrong.

### 2. `coding-dev.yaml` was not natively integrated

`ensure_notebook` had to do:

1. `lab-notebook init` (writes default `research-notebook` schema)
2. `cp coding-dev.yaml .lnb/schema.yaml` (overwrite)
3. `lab-notebook rebuild` (re-sync SQLite type vocabulary)

…plus a fallback branch that emitted a warning and silently degraded to `research-notebook` if the schema file was missing. Clunky because `lab-notebook` had no way to load an external schema at init time.

This got fixed upstream: lab-notebook [#15](https://github.com/carbonscott/lab-notebook/issues/15) added `--template-path PATH` (commit `06af5e8`). A separate commit `ed34b5a` made `.lnb` always a subfolder of the passed path, retiring the path-doubling quirk mentioned in problem (1).

### 3. Flat repo structure hid the mode boundary

All seven files (`ralph.sh`, `ralph-prep.sh`, `ralph-lib.sh`, `RALPH-CC.md`, `PROMPT.md`, `coding-dev.yaml`, `tasks.json.example`) sat at repo root. Nothing told a reader which belonged to the headless runner, which to the Claude Code session runner, and which were shared between them.

## Proposal

Restructure into three directories:

```
ralph-wiggum-lnb/
├── shared/
│   ├── PROMPT.md
│   ├── ralph-lib.sh
│   ├── coding-dev.yaml
│   └── tasks.json.example
├── cc-headless/
│   └── ralph.sh
└── cc/
    ├── RALPH-CC.md
    └── ralph-prep.sh
```

Naming: `cc` for the Claude-Code-session mode, `cc-headless` for the `claude -p` mode. Picking `cc-headless` rather than `cc` + `non-headless` makes the default-ness of non-headless implicit.

Alongside the restructure:

- **Rewrite `ensure_notebook`** to use `lab-notebook init --template-path "$SHARED_DIR/coding-dev.yaml"`. Drop the cp + rebuild + warn-on-missing branches and the long path-doubling comment.
- **Drop the manual bootstrap from README.** `ensure_notebook` handles it on first iteration.
- **Zero-copy invocation model.** Users only need `PROMPT.md` + `tasks.json` in their project dir. The runners are invoked by absolute path from the repo (`~/codes/ralph-wiggum-lnb/cc-headless/ralph.sh` or `follow ~/codes/ralph-wiggum-lnb/cc/RALPH-CC.md` in a Claude Code session). No more copying `ralph-prep.sh` / `ralph-lib.sh` / `coding-dev.yaml` into every project.

## Scope

- Seven `git mv`s (history preserved).
- Shell-script edits: add `SHARED_DIR="$SCRIPT_DIR/../shared"` to both runners, update the sourced-lib path, update two error-hint messages.
- Rewrite `ensure_notebook` (net -20 lines).
- Update `README.md`, `cc/RALPH-CC.md`, `tests/setup-sandbox.sh`, `tests/README.md` to reference new paths and shrink the prerequisites.

PR incoming.

carbonscott:
Closes #10.

## Summary

- **Restructure** into `shared/`, `cc-headless/`, `cc/` so the two-runner boundary is obvious at a glance.
- **Simplify `ensure_notebook`** to use `lab-notebook init --template-path "$SHARED_DIR/coding-dev.yaml"` (upstream flag landed in lab-notebook `06af5e8`), dropping the old cp + rebuild + warn-on-missing dance and the now-obsolete path-doubling comment.
- **Drop the manual `.lnb` bootstrap** from the README — it was both broken (triggered the retired path-doubling quirk) and redundant with `ensure_notebook`.
- **Zero-copy invocation model**: a user's project dir now contains only `PROMPT.md` + `tasks.json`. Runners, the shared lib, and the notebook schema all stay in this repo and are invoked or sourced by absolute path. No more copying `ralph-prep.sh` / `ralph-lib.sh` / `coding-dev.yaml` into every project.

## Layout diff

Before (flat):
```
ralph.sh  ralph-prep.sh  ralph-lib.sh  RALPH-CC.md  PROMPT.md  coding-dev.yaml  tasks.json.example
```

After:
```
shared/
  PROMPT.md
  ralph-lib.sh
  coding-dev.yaml
  tasks.json.example
cc-headless/
  ralph.sh
cc/
  RALPH-CC.md
  ralph-prep.sh
```

Naming: `cc-headless` (rather than `cc` + `non-headless`) makes the default-ness of the in-session mode implicit.

## Invocation model change

**Old** — users copied 4–5 files into their project:
```bash
cp ~/codes/ralph-wiggum-lnb/{RALPH-CC.md,ralph-prep.sh,ralph-lib.sh,coding-dev.yaml,PROMPT.md} .
chmod +x ralph-prep.sh
```

**New** — users copy 2 files, invoke runners by absolute path:
```bash
cp ~/codes/ralph-wiggum-lnb/shared/PROMPT.md .
cp ~/codes/ralph-wiggum-lnb/shared/tasks.json.example tasks.json
# headless:
~/codes/ralph-wiggum-lnb/cc-headless/ralph.sh --max-iterations 5
# or in a Claude Code chat:
# follow ~/codes/ralph-wiggum-lnb/cc/RALPH-CC.md, max-iterations 5
```

## Test plan

Already executed locally against a fresh repo checkout — listed here so a reviewer can re-run.

- [x] `bash -n` on all four shell files — no syntax errors.
- [x] `cc-headless/ralph.sh --help` and `cc/ralph-prep.sh --help` parse arguments and print usage.
- [x] **End-to-end bootstrap in a fresh temp dir:**
  ```bash
  TMP=$(mktemp -d) && cd $TMP
  cp ~/codes/ralph-wiggum-lnb/shared/PROMPT.md .
  cp ~/codes/ralph-wiggum-lnb/tests/tasks.json .
  ~/codes/ralph-wiggum-lnb/cc/ralph-prep.sh --iteration 1 --max-iterations 1 > /dev/null
  diff $TMP/.lnb/schema.yaml ~/codes/ralph-wiggum-lnb/shared/coding-dev.yaml   # no output = match
  LAB_NOTEBOOK_DIR=$TMP/.lnb lab-notebook sql "SELECT DISTINCT type FROM entries"
  ```
  Schema diff is empty (byte-match), `type` column returns `start` — confirms the coding-dev schema is active, not the research-notebook default.
- [x] **Non-default notebook dir** (`--notebook mynb`): `mynb/` is created with the coding-dev schema, `.lnb/` is absent (rename succeeded), `.lnb.env` contains `LAB_NOTEBOOK_DIR=<abs>/mynb`.
- [x] **`tests/setup-sandbox.sh`**: produces a sandbox with exactly `PROMPT.md` (symlinked to `shared/PROMPT.md`) + `tasks.json`. Printed instructions reference `$REPO_DIR/cc-headless/ralph.sh` and `$REPO_DIR/cc/RALPH-CC.md` with absolute paths resolved.
- [ ] **End-to-end smoke via headless runner** — left for reviewer to run against the sandbox (requires a live `claude` CLI session that's costly to exercise in CI).
- [ ] **End-to-end smoke via Claude Code session** — same.

## Follow-ups not in this PR

- `lab-notebook` issue [#15](https://github.com/carbonscott/lab-notebook/issues/15) proposed `--template` path-fallback; the landed implementation chose `--template-path` as a separate flag instead. That's cleaner and this PR uses it. No action needed.


shared/ralph-lib.sh
```diff
@@ -38,36 +41,20 @@ archive_previous_run() {
...
-        # `.lnb/` in cwd. Passing a positional arg triggers a
-        # path-doubling quirk that nests the real notebook at
-        # `<dir>/.lnb`, leaving any pre-placed schema.yaml as a ghost
-        # file. Stick with the default, then rename if the caller
-        # wanted a non-default notebook name.
-        lab-notebook init >/dev/null 2>&1 || true
+        # lab-notebook init always creates a `.lnb` subfolder in cwd and
+        # loads the schema from --template-path directly, so no separate
+        # cp + rebuild step is needed.
+        lab-notebook init --template-path "$SHARED_DIR/coding-dev.yaml" >/dev/null
```
carbonscott, line 47: Good call dropping `|| true` here. The outer `if [[ ! -d "$NOTEBOOK_DIR" ]]` guard keeps this idempotent across re-runs — `init` only fires when the notebook dir is absent — so silently swallowing errors only masked real failures (bad schema path, missing `lab-notebook` binary, permission issue). Failing loudly is the right behavior.

FYI one narrow edge case this PR does not address (and doesn't need to): if someone previously ran with `--notebook .lnb` and now re-runs with `--notebook mynb`, the guard on `mynb` passes but there's an existing `.lnb/` in cwd. Depending on how `lab-notebook init` behaves when `.lnb/` already exists, you could get a clobber or a hard fail. Pre-existing, not a regression.

shared/ralph-lib.sh
```diff
@@ -38,36 +41,20 @@ archive_previous_run() {
...
 
+        # Rename if the caller wants a non-default notebook dir, and
+        # patch .lnb.env so lab-notebook calls still find the notebook.
         if [[ "$NOTEBOOK_DIR" != ".lnb" && -d ".lnb" && ! -d "$NOTEBOOK_DIR" ]]; then
             mv .lnb "$NOTEBOOK_DIR"
-            # Keep .lnb.env pointing at the renamed notebook so
-            # lab-notebook calls without an explicit LAB_NOTEBOOK_DIR
-            # env var still find it.
             if [[ -f .lnb.env ]]; then
                 sed -i.bak "s|/\\.lnb$|/$NOTEBOOK_DIR|" .lnb.env && rm -f .lnb.env.bak
```
carbonscott, line 54: Nit / Optional: the sed pattern is safe as-is — the `$` anchor plus the leading `/` require the match to be `.../ .lnb` at end-of-line, so a project path like `/home/user/foo.lnb/project` will not be touched and a notebook path like `/home/user/foo.lnb/project/.lnb` rewrites correctly.

One defensive improvement worth considering: this code assumes `.lnb.env` contains a single line matching `…/\.lnb$`. If lab-notebook ever changes its env file format (trailing slash, comments, multiple vars, etc.), the sed silently no-ops and `lab-notebook` calls that rely on `.lnb.env` stop finding the renamed notebook. A one-line `grep -q '/\.lnb$' .lnb.env || echo "warning: unexpected .lnb.env format, skipping rename" >&2` would catch a format drift loudly.

cc-headless/ralph.sh
```diff
@@ -12,6 +12,7 @@ COMPLETION_PROMISE="DONE"
 ALL_DONE_PROMISE="ALL_DONE"
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SHARED_DIR="$SCRIPT_DIR/../shared"
```
carbonscott, line 23: Optional: `SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"` does not follow symlinks *to the script file itself*. If a user symlinks `~/bin/ralph -> .../cc-headless/ralph.sh` and runs `ralph`, `BASH_SOURCE[0]` is the symlink path, `dirname` yields `~/bin`, and `SHARED_DIR` resolves to `~/shared` — wrong.

This was also true before this PR (`$SCRIPT_DIR/PROMPT.md` would have been equally broken), so it's not a regression. But since the new invocation model encourages "run by absolute path from the repo," users are more likely to want to `ln -s` this into `$PATH`. Two options, whichever you prefer:

1. Resolve the symlink explicitly: `SCRIPT_DIR="$(cd "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")" && pwd)"` (GNU `readlink -f`; `python3 -c 'import os,sys;print(os.path.realpath(sys.argv[1]))' "${BASH_SOURCE[0]}"` is the portable fallback).
2. Document that users should symlink a wrapper script, not the runner itself.

Non-blocking — the doc-recommended invocation (absolute path) already avoids this entirely.

cc/ralph-prep.sh
```diff
@@ -16,6 +16,7 @@ MAX_ITERATIONS=""
 LAST_BRANCH_FILE=".ralph-last-branch"
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SHARED_DIR="$SCRIPT_DIR/../shared"
```
carbonscott, line 27: Same symlink footgun as `cc-headless/ralph.sh:15` — this `SCRIPT_DIR` pattern doesn't resolve a symlink pointing at the script file. Not a regression, just worth knowing for any future user who symlinks into `$PATH`.

cc/RALPH-CC.md
```diff
@@ -68,14 +68,15 @@ For `i` in `1..max-iterations`:
 1. **Build the prompt.** Call:
 
    ```
-   Bash("./ralph-prep.sh --iteration i --max-iterations N --task-file <task-file>")
+   Bash("~/codes/ralph-wiggum-lnb/cc/ralph-prep.sh --iteration i --max-iterations N --task-file <task-file>")
```
carbonscott: The `~` here will expand correctly when the Claude Code `Bash` tool forwards the command to `bash -c` — unquoted leading `~` gets tilde-expanded by the shell, so this works as written. Not a bug.

However, this is a driver doc that an LLM executes mechanically, and LLMs sometimes wrap arguments in single quotes "for safety," which would suppress tilde expansion (`'~/codes/...'` becomes a literal path). Two cheap hardenings:

1. Use `$HOME/codes/ralph-wiggum-lnb/cc/ralph-prep.sh` — `$HOME` expands inside double quotes and is less likely to be quoted away.
2. Or, since line 78–79 already says "if the repo lives somewhere else, substitute your checkout path," you could instruct the main session to resolve the path once during Setup (e.g. `Bash("readlink -f ~/codes/ralph-wiggum-lnb/cc/ralph-prep.sh")`) and reuse the absolute result for every iteration.

Non-blocking — current form works in the common case.

cc/RALPH-CC.md
```diff
@@ -30,14 +30,14 @@ Defaults when unspecified:
...
-3. `ralph-prep.sh` must be executable (`chmod +x ralph-prep.sh`).
-4. `jq` and `lab-notebook` must be on `$PATH` (same as for `ralph.sh`).
+2. `PROMPT.md` and `tasks.json` must exist in the current directory.
+   Copy them from the repo: `cp ~/codes/ralph-wiggum-lnb/shared/PROMPT.md .`
+   and `cp ~/codes/ralph-wiggum-lnb/shared/tasks.json.example tasks.json`,
+   then edit `tasks.json` to describe your stories. Nothing else needs
+   to live in the project dir — `ralph-prep.sh`, `ralph-lib.sh`, and
+   `coding-dev.yaml` all stay in the repo and are invoked/sourced by
+   absolute path.
+3. `jq` and `lab-notebook` must be on `$PATH` (same as for `ralph.sh`).
```
carbonscott, line 40: Nice shrinking of the prerequisites list. Dropping `ralph-prep.sh`, `ralph-lib.sh`, `coding-dev.yaml` from "must exist in the current directory" is the main usability win of this refactor.

tests/setup-sandbox.sh
```diff
@@ -6,19 +6,19 @@ SANDBOX="${TMPDIR:-/tmp}/ralph-smoke-$(date +%Y%m%d-%H%M%S)"
...
 mkdir -p "$SANDBOX"
 
-for f in PROMPT.md ralph.sh ralph-prep.sh ralph-lib.sh RALPH-CC.md coding-dev.yaml; do
-    ln -s "$REPO_DIR/$f" "$SANDBOX/$f"
-done
-
+# The sandbox mimics a user's project dir: only PROMPT.md + tasks.json.
+# Runners are invoked by absolute path from $REPO_DIR, exercising the
+# zero-copy invocation model that cc-headless and cc modes both use.
+ln -s "$REPO_DIR/shared/PROMPT.md" "$SANDBOX/PROMPT.md"
```
carbonscott, line 12: Good call symlinking only `PROMPT.md` and copying `tasks.json`. The `PROMPT.md` symlink means uncommitted edits to the template take effect in the smoke test (which is what `tests/README.md` Gotchas already describes), and the `tasks.json` copy isolates the sandbox from any accidental fixture mutation. The two different mechanisms are justified — worth noting if someone asks later.

README.md
```diff
@@ -12,34 +12,32 @@ structured notebook logging (queryable history, pattern discovery). See also
...
-cp ~/codes/ralph-wiggum-lnb/PROMPT.md .
-cp ~/codes/ralph-wiggum-lnb/tasks.json.example tasks.json
+cp ~/codes/ralph-wiggum-lnb/shared/PROMPT.md .
+cp ~/codes/ralph-wiggum-lnb/shared/tasks.json.example tasks.json
 # Edit tasks.json with your stories
-
-# Initialize notebook with coding schema
-mkdir -p .lnb && cp ~/codes/ralph-wiggum-lnb/coding-dev.yaml .lnb/schema.yaml
-lab-notebook init .lnb
 ```
```
carbonscott, line 19: The "runner auto-initializes `.lnb/` on the first iteration" note is exactly the right level of detail — tells the user "don't bother" without dragging them into the `ensure_notebook` internals. Good doc edit.

carbonscott (commented):
## Summary

Clean, well-scoped refactor. The three-directory split (`shared/`, `cc-headless/`, `cc/`) makes the two-runner boundary obvious at a glance, and the zero-copy invocation model is a real usability win — shrinking a project's ralph footprint from 4–5 files to 2 is exactly the kind of simplification that tends to get appreciated six months later. The `ensure_notebook` rewrite is a pleasure to read: 20 lines replaced with ~10, and the deleted `cp`+`rebuild`+`warn-on-missing` dance was brittle enough that losing it is a net improvement.

Recommendation: **COMMENT**. Nothing here is blocking — everything I'd flag is either a pre-existing behavior this PR perpetuates, or a minor doc ambiguity. The refactor definitely improves overall code health.

## Walkthrough of the things you asked me to look at

### 1. Removing `|| true` from `lab-notebook init` — correct call

Idempotency is preserved by the outer `if [[ ! -d "$NOTEBOOK_DIR" ]]` guard. A second run inside an initialized notebook dir skips the whole block, so `init` never fires twice. Failing loudly on a bad schema path is the right trade-off.

One narrow edge case worth knowing about (not a regression — pre-existing): if a user previously ran with `--notebook .lnb` and now re-runs with `--notebook mynb`, the outer guard (`! -d mynb`) passes, and `lab-notebook init` is called with an already-present `.lnb/` in cwd. Depending on lab-notebook's init semantics, this may either clobber or fail. Not new in this PR, not worth holding the merge.

### 2. `SHARED_DIR` resolution via `$SCRIPT_DIR/../shared` — works for direct invocation, breaks under symlinks

The pattern `SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"` does **not** resolve symlinks to the script file itself. If a user symlinks `~/bin/ralph -> ~/codes/ralph-wiggum-lnb/cc-headless/ralph.sh`, then `BASH_SOURCE[0]=~/bin/ralph`, `dirname=~/bin`, `cd ~/bin && pwd=~/bin` (the `cd` only follows symlinks when the argument is a symlink to a *directory*; here it's a file, so we never cd through it), and `SHARED_DIR=~/bin/../shared=~/shared` — wrong.

This was already true in the old layout (`$SCRIPT_DIR/PROMPT.md` would have been equally broken), so this PR does not regress anything. But since the PR is explicitly pivoting the docs toward "invoke the runner by absolute path from the repo," it's worth either:
- resolving symlinks properly with `readlink -f` / a canonical-path helper, or
- adding a short comment warning that users who symlink the runner into `$PATH` should point the symlink at a wrapper script, not the runner itself.

I'd make this a follow-up, not a blocker. See inline on `cc-headless/ralph.sh:15`.

### 3. `.lnb.env` rename sed — safe, you called it right

Pattern: `s|/\.lnb$|/$NOTEBOOK_DIR|`. The `$` anchor + leading `/` together require the match to be `…/.lnb` at end-of-line, so `/home/user/foo.lnb/project/.lnb` matches (rewrites the trailing `.lnb`), but `/home/user/foo.lnb/project` does **not** (no `.lnb` tail), and `/home/user/foo.lnb` does **not** (no `/` immediately before `.lnb`). The anchor does the work. Good call.

Nit: since `.lnb.env` is generated by `lab-notebook init` on the same invocation and immediately consumed here, its format is effectively an upstream contract. If lab-notebook ever starts writing an absolute path with a trailing slash, or comments, or multiple entries, this sed breaks silently. A one-line assertion like `grep -q '/\.lnb$' .lnb.env` before the `sed` would catch a format drift loudly. Optional.

### 4. `~` expansion in `Bash()` tool calls — works, but worth a sentence

The Claude Code `Bash` tool forwards the command string to a bash shell (`bash -c`), which performs tilde expansion on unquoted leading `~`. So `Bash("~/codes/ralph-wiggum-lnb/cc/ralph-prep.sh --iteration i ...")` does expand correctly. Not a bug.

However, the phrasing on line 78–79 ("If the repo lives somewhere other than `~/codes/ralph-wiggum-lnb`, substitute your checkout path") is a little thin for a driver doc that's executed mechanically by an LLM. If the session happens to wrap the path in single quotes, tilde expansion silently fails. Consider either (a) showing `$HOME/codes/...` which expands in more quoting contexts, or (b) instructing the main session to use an absolute path explicitly. See inline on `cc/RALPH-CC.md:71`.

### 5. `tests/README.md` Gotchas — consistent with the new setup-sandbox

The Gotchas section now says "symlinks `PROMPT.md` from `shared/` and runs scripts live from the repo by absolute path" which matches `setup-sandbox.sh` exactly. Consistent.

### 6. Dead references to flat paths — none

Searched the whole repo for `ralph.sh`, `ralph-prep.sh`, `ralph-lib.sh`, `RALPH-CC.md`, `coding-dev.yaml`, `tasks.json.example` — every reference either (a) uses the new prefixed path, (b) is a bare filename inside a runner where the file lives alongside the script (e.g. `PROMPT.md` default), or (c) is prose inside a code comment or doc that legitimately refers to the file by short name. No stale path references.

## Things I liked

- The header comment block in `shared/ralph-lib.sh` now explicitly lists the contract (`SCRIPT_DIR`, `SHARED_DIR`, `PROMPT_FILE`, …) and tells callers how to set `SHARED_DIR`. Good defensive doc for a sourced file.
- Deleting the `warn-on-missing coding-dev.yaml` branch in `ensure_notebook` — that warning was user-hostile because by the time you saw it, the notebook was already initialized with the wrong schema and the only fix was to blow away `.lnb/`. Failing at init is much better.
- The `Relationship to shared/ralph-lib.sh` section at the bottom of `cc/RALPH-CC.md` is the right place to remind future editors that ralph-lib changes affect both runners.
- The `cc-headless/` naming choice is nice — it implicitly communicates that the in-session mode is the "default" mental model, not an exotic variant.

## Style / nits (non-blocking)

- `shared/ralph-lib.sh:1` still says `# Shared helpers for ralph.sh and ralph-prep.sh.` It's tiny, but `cc-headless/ralph.sh` and `cc/ralph-prep.sh` would be more precise now.
- `cc/ralph-prep.sh:4` — the header comment says "for RALPH-CC.md" without the `cc/` prefix. Trivial.

PR #13, opened 2026-04-15
feat: install.sh + /ralph-lnb skill for ergonomic invocation

### LINKED ISSUE #12: Install script + /ralph-lnb skill for ergonomic invocation
## Problem

PR #11 restructured the repo into `shared/` / `cc-headless/` / `cc/` and moved users to a zero-copy invocation model. That fixed file sprawl but made the commands themselves verbose:

- **Headless**: `/Users/cwang31/codes/ralph-wiggum-lnb/cc-headless/ralph.sh --max-iterations 3`
- **CC session**: `follow /Users/cwang31/codes/ralph-wiggum-lnb/cc/RALPH-CC.md, max-iterations 3`

Two problems in one:
1. Both invocations embed an absolute path the user has to remember or type.
2. Repo location is load-bearing — moving or re-cloning the repo breaks any shortcuts the user builds around those paths.

The ergonomic target: after `git clone <url> ~/wherever`, run one command, then type `ralph --max-iterations 3` (headless) or `/ralph-lnb max-iterations 3` (CC chat). Location of the repo on disk stops mattering.

## Proposal

Ship an `install.sh` at the repo root that creates two install artifacts:

1. **`~/.local/bin/ralph`** → symlink to `cc-headless/ralph.sh`. Thanks to the symlink-walking we added in commit `791a045`, `SHARED_DIR` still resolves correctly when invoked via the symlink.
2. **`~/.claude/skills/ralph-lnb/SKILL.md`** → rendered from a `skill/SKILL.md.template` by substituting a `@@RALPH_REPO@@` placeholder with the repo's absolute path at install time.

**Skill, not slash command.** Anthropic [merged slash commands into skills](https://code.claude.com/docs/en/skills) in January 2026. A skill with `name: ralph-lnb` in its frontmatter auto-registers as `/ralph-lnb` — same UX, but the format is future-proof.

**Name:** `/ralph-lnb` (not `/ralph`) to avoid collision with the common short name and to identify this specific ralph variant (lab-notebook-integrated).

### Installer design

- Resolves its own path through symlinks (same pattern as the runners), so `install.sh` works when called via a symlink or wrapper.
- Idempotent: safe to re-run. Recognizes artifacts it owns and refreshes them.
- Refuses to clobber unmanaged files (non-symlinks at the bin target, or skill files without the manage marker) unless `--force` is passed.
- Prints a post-install summary with usage examples and warns if `$BIN_DIR` isn't on `$PATH`.
- `RALPH_BIN_DIR=/usr/local/bin ./install.sh` overrides the bin location for people who want a sudo-style install.
- Mirror `uninstall.sh` removes only managed artifacts, cleans up empty dirs.

### Skill template

A parallel file to `cc/RALPH-CC.md`, not a templated transform of it. The two serve different audiences (a human reading the driver doc vs. Claude executing a skill trigger), and trying to make one file work for both would need conditional logic. Diff drift between the two is reviewable.

The skill template:
- Frontmatter (`name: ralph-lnb`, `description:`, `user-invocable: true`, `argument-hint:`).
- Marker comment on line 1 of the body so `uninstall.sh` can detect managed files.
- Parses `$ARGUMENTS` at the top, dispatches to Setup → Loop → Post-loop following the same procedure as `RALPH-CC.md`.
- All `ralph-prep.sh` paths use `@@RALPH_REPO@@` → substituted at install time.
- Ends with `$ARGUMENTS` on the last line (convention from other skills on this machine).

### What stays unchanged

- `cc-headless/ralph.sh`, `cc/ralph-prep.sh`, `shared/*`, `cc/RALPH-CC.md`, `tests/*`. No existing behavior is modified. Users who prefer not to install still get the absolute-path invocation documented in README Quick Start.

## Scope

- New: `install.sh`, `uninstall.sh`, `skill/SKILL.md.template`.
- Modified: `README.md` (new Install section above Quick Start).
- Zero changes to existing runtime scripts.

## Delivery

Separate PR (not folded into #11) but **based on** `refactor/cc-headless-layout` — stacked PR pattern. GitHub auto-rebases onto `master` when #11 merges.

PR incoming.

carbonscott:
Closes #12.

**Stacked on #11** (`refactor/cc-headless-layout`) — the base of this PR is `refactor/cc-headless-layout`, not `master`. GitHub will auto-rebase to `master` when #11 merges.

## Summary

- **`install.sh`**: symlink-safe, idempotent installer that creates two artifacts and nothing else:
  - `~/.local/bin/ralph` → symlink to `cc-headless/ralph.sh`
  - `~/.claude/skills/ralph-lnb/SKILL.md` → rendered from `skill/SKILL.md.template` with the absolute repo path baked in
- **`uninstall.sh`**: mirror; only touches artifacts marked as managed.
- **`skill/SKILL.md.template`**: the `/ralph-lnb` skill body. Parallel to `cc/RALPH-CC.md` rather than templated from it, because the two serve different audiences (Claude in a skill trigger vs. a human reader).
- **README**: new Install section above Quick Start. Absolute-path invocations stay documented as the no-install fallback.

## Zero changes to existing runtime

No edits to `cc-headless/ralph.sh`, `cc/ralph-prep.sh`, `cc/RALPH-CC.md`, `shared/*`, or `tests/*`. Users who prefer not to install still use the same absolute-path invocations from PR #11.

## Design notes

- **Skills, not slash commands.** Anthropic merged slash commands into skills in January 2026. A skill with `name: ralph-lnb` auto-registers as `/ralph-lnb`. Ships with `user-invocable: true` + `argument-hint` frontmatter fields.
- **Name `/ralph-lnb`**, not `/ralph` — avoids collision with the common short name and identifies this lab-notebook-integrated variant.
- **Safety.** `install.sh` refuses to overwrite non-symlink files at the bin target or skill files without the `Managed by ralph-wiggum-lnb install.sh` marker, unless `--force` is passed. `uninstall.sh` has the same guardrails in reverse.
- **Idempotence.** Re-running `install.sh` recognizes its own artifacts and refreshes them. Re-running after moving the repo is the intended way to re-point paths at the new location.
- **Path resolution.** `install.sh` uses the same `BASH_SOURCE`-walking pattern that commit `791a045` added to the runners, so it works correctly when called via a symlink.

## Test plan

Already executed locally. Listed here for reviewer repro.

### 1. Fresh install (clean state)
```bash
rm -f ~/.local/bin/ralph
rm -rf ~/.claude/skills/ralph-lnb
./install.sh
```
- [x] `readlink ~/.local/bin/ralph` → `$REPO/cc-headless/ralph.sh`
- [x] `~/.claude/skills/ralph-lnb/SKILL.md` exists
- [x] `grep -c '@@RALPH_REPO@@' ~/.claude/skills/ralph-lnb/SKILL.md` → 0
- [x] `grep -c "$REPO" ~/.claude/skills/ralph-lnb/SKILL.md` → 6 (all path references substituted)
- [x] `grep -q 'Managed by ralph-wiggum-lnb install.sh' ~/.claude/skills/ralph-lnb/SKILL.md` → present (marker for uninstall detection)
- [x] `ralph --help` works from the symlink

### 2. Idempotence
```bash
./install.sh  # second run
```
- [x] No errors. Refresh-in-place messaging. Both artifacts unchanged.

### 3. Repo-move simulation (without `--force`)
```bash
cp -R $REPO $MOVED
$MOVED/install.sh
```
- [x] Bin symlink **refused** because it points at the original repo (not this one) — safety check triggers.
- [x] Skill was managed, so it gets refreshed to point at `$MOVED`.

### 4. Repo-move simulation (with `--force`)
```bash
$MOVED/install.sh --force
```
- [x] `Overwriting /Users/.../ralph (was: <old-path>)` — explicit "what changed" message.
- [x] Both artifacts now point at `$MOVED`.

### 5. Non-symlink safety
```bash
touch ~/.local/bin/ralph
./install.sh          # should warn and skip
./install.sh --force  # should overwrite with "Overwriting non-symlink file"
```
- [x] Without `--force`: bin skipped with warning, skill still installed (independent paths).
- [x] With `--force`: bin overwritten with clear message.

### 6. Uninstall
```bash
./uninstall.sh
```
- [x] Both artifacts removed. Empty `ralph-lnb/` dir cleaned up.

### 7. Re-uninstall (no-op)
```bash
./uninstall.sh
```
- [x] Prints `"Nothing to remove."`

### 8. Syntax
- [x] `bash -n install.sh uninstall.sh` passes.

### 9. Manual verification
- [ ] Reviewer: restart Claude Code, confirm `/ralph-lnb` appears in the slash-command autocomplete. Skills are cached per session, so newly installed skills require a session restart to show up reliably.

## Out of scope

- No refactor of `cc/RALPH-CC.md` to dedup with the skill template. Parallel files, diff-reviewable.
- No bash tab completion for `ralph`.
- No `--purge` flag in uninstall (user data like `archive/` is never touched).
- No Homebrew/apt packaging. Git-clone-and-go is the shipping model.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,9 @@ Two install artifacts:
 
 - `~/.local/bin/ralph` → symlink to the headless runner. Put `~/.local/bin`
   on `$PATH` if it isn't already (`install.sh` warns you if not).
-- `~/.claude/skills/ralph-lnb/SKILL.md` → the `/ralph-lnb` slash command
-  for Claude Code chats.
+- `~/.claude/skills/ralph-lnb/SKILL.md` → the `ralph-lnb` skill. Claude
+  Code exposes user-invocable skills as slash commands, so it appears
+  as `/ralph-lnb` in chat.
 
 After install, the invocations become:
 
@@ -31,7 +32,8 @@ After install, the invocations become:
 `install.sh` is idempotent. Re-run it after moving or re-cloning the
 repo — it rewrites the skill with the new path. Override the bin
 location with `RALPH_BIN_DIR=/usr/local/bin ./install.sh`. Undo with
-`./uninstall.sh`.
+`./uninstall.sh` — if you overrode `RALPH_BIN_DIR` on install, pass
+the same value on uninstall so it can find the symlink to remove.
 
 If you prefer not to install, the absolute-path invocations in the
 Quick Start below still work.
--- a/README.md
+++ b/README.md
@@ -8,6 +8,34 @@ Based on the [Ralph Wiggum technique](https://ghuntley.com/ralph/) with
 structured notebook logging (queryable history, pattern discovery). See also
 [Effective Harnesses for Long-Running Agents](https://www.anthropic.com/engineering/effective-harnesses-for-long-running-agents).
 
+## Install
+
+```bash
+git clone https://github.com/carbonscott/ralph-wiggum-lnb ~/codes/ralph-wiggum-lnb
+~/codes/ralph-wiggum-lnb/install.sh
+```
+
+Two install artifacts:
+
+- `~/.local/bin/ralph` → symlink to the headless runner. Put `~/.local/bin`
+  on `$PATH` if it isn't already (`install.sh` warns you if not).
+- `~/.claude/skills/ralph-lnb/SKILL.md` → the `/ralph-lnb` slash command
+  for Claude Code chats.
+
+After install, the invocations become:
+
+- **Headless**: `ralph --max-iterations 3`
+- **Claude Code chat**: `/ralph-lnb max-iterations 3` (restart any
+  running sessions — skills load at session start)
+
+`install.sh` is idempotent. Re-run it after moving or re-cloning the
+repo — it rewrites the skill with the new path. Override the bin
+location with `RALPH_BIN_DIR=/usr/local/bin ./install.sh`. Undo with
+`./uninstall.sh`.
+
+If you prefer not to install, the absolute-path invocations in the
+Quick Start below still work.
+
 ## Quick Start
 
 ```bash
--- a/install.sh
+++ b/install.sh
@@ -115,8 +115,11 @@ fi
 
 if [[ $install_skill -eq 1 ]]; then
     # Render template: substitute @@RALPH_REPO@@ with the absolute repo path.
-    # Use | as sed delimiter since paths contain / — REPO shouldn't contain |.
-    sed "s|@@RALPH_REPO@@|$REPO|g" "$TEMPLATE" > "$SKILL_TARGET"
+    # Escape \, |, & in $REPO — those are the three characters sed treats
+    # specially in the replacement string. Backslash must go first so the
+    # subsequent substitutions don't double-escape.
+    REPO_ESCAPED=$(printf '%s\n' "$REPO" | sed 's/[\\|&]/\\&/g')
+    sed "s|@@RALPH_REPO@@|$REPO_ESCAPED|g" "$TEMPLATE" > "$SKILL_TARGET"
     echo "Installed: $SKILL_TARGET (rendered with $REPO)"
 fi
 
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,151 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Install ralph-wiggum-lnb ergonomic entry points:
+#   1. Symlink this repo's cc-headless/ralph.sh into $BIN_DIR (default: ~/.local/bin)
+#      so `ralph` runs from any cwd.
+#   2. Render skill/SKILL.md.template -> ~/.claude/skills/ralph-lnb/SKILL.md
+#      with @@RALPH_REPO@@ substituted, so `/ralph-lnb` works in Claude Code chats.
+#
+# Idempotent: safe to re-run after moving or re-cloning the repo. Pass --force
+# to overwrite install artifacts that don't belong to ralph-wiggum-lnb.
+
+# --- Resolve this script's real location (symlink-safe) ---
+SOURCE="${BASH_SOURCE[0]}"
+while [ -L "$SOURCE" ]; do
+    DIR="$(cd "$(dirname "$SOURCE")" && pwd)"
+    SOURCE="$(readlink "$SOURCE")"
+    [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE"
+done
+REPO="$(cd "$(dirname "$SOURCE")" && pwd)"
+
+# --- Parse args ---
+FORCE=0
+for arg in "$@"; do
+    case "$arg" in
+        --force) FORCE=1 ;;
+        -h|--help)
+            cat <<'EOF'
+Usage: install.sh [--force]
+
+Installs ralph-wiggum-lnb entry points. Run from a fresh git clone:
+
+    git clone <url> ~/codes/ralph-wiggum-lnb
+    ~/codes/ralph-wiggum-lnb/install.sh
+
+Environment variables:
+  RALPH_BIN_DIR   Where to put the `ralph` symlink. Default: ~/.local/bin
+
+Options:
+  --force         Overwrite install artifacts that don't belong to this repo
+  -h, --help      Show this help
+
+Run ./uninstall.sh from the repo to undo.
+EOF
+            exit 0 ;;
+        *) echo "Unknown option: $arg" >&2; exit 2 ;;
+    esac
+done
+
+BIN_DIR="${RALPH_BIN_DIR:-$HOME/.local/bin}"
+SKILL_DIR="$HOME/.claude/skills/ralph-lnb"
+BIN_TARGET="$BIN_DIR/ralph"
+SKILL_TARGET="$SKILL_DIR/SKILL.md"
+TEMPLATE="$REPO/skill/SKILL.md.template"
+MARKER="Managed by ralph-wiggum-lnb install.sh"
+
+# --- Preflight: template must exist ---
+if [[ ! -f "$TEMPLATE" ]]; then
+    echo "Error: template not found at $TEMPLATE" >&2
+    echo "Is this a complete ralph-wiggum-lnb checkout?" >&2
+    exit 1
+fi
+
+# --- Install 1: symlink `ralph` into BIN_DIR ---
+mkdir -p "$BIN_DIR"
+
+install_bin=1
+if [[ -e "$BIN_TARGET" || -L "$BIN_TARGET" ]]; then
+    if [[ -L "$BIN_TARGET" ]]; then
+        existing="$(readlink "$BIN_TARGET")"
+        if [[ "$existing" == "$REPO/cc-headless/ralph.sh" ]]; then
+            echo "Symlink $BIN_TARGET already points at this repo — refreshing."
+        else
+            if [[ $FORCE -eq 1 ]]; then
+                echo "Overwriting $BIN_TARGET (was: $existing)"
+            else
+                echo "Warning: $BIN_TARGET is a symlink to $existing, not this repo." >&2
+                echo "         Skipping. Re-run with --force to overwrite." >&2
+                install_bin=0
+            fi
+        fi
+    else
+        if [[ $FORCE -eq 1 ]]; then
+            echo "Overwriting non-symlink file at $BIN_TARGET"
+        else
+            echo "Warning: $BIN_TARGET exists and is not a symlink." >&2
+            echo "         Skipping. Re-run with --force to overwrite." >&2
+            install_bin=0
+        fi
+    fi
+fi
+
+if [[ $install_bin -eq 1 ]]; then
+    ln -sf "$REPO/cc-headless/ralph.sh" "$BIN_TARGET"
+    echo "Installed: $BIN_TARGET -> $REPO/cc-headless/ralph.sh"
+fi
+
+# --- Install 2: render skill template ---
+mkdir -p "$SKILL_DIR"
+
+install_skill=1
+if [[ -f "$SKILL_TARGET" ]]; then
+    if grep -q "$MARKER" "$SKILL_TARGET"; then
+        echo "Skill $SKILL_TARGET is managed by install.sh — refreshing."
+    else
+        if [[ $FORCE -eq 1 ]]; then
+            echo "Overwriting unmanaged skill at $SKILL_TARGET"
+        else
+            echo "Warning: $SKILL_TARGET exists and is not managed by this installer." >&2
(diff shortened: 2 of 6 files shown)


install.sh
```diff
@@ -0,0 +1,151 @@
...
+            echo "         Skipping. Re-run with --force to overwrite." >&2
+            install_skill=0
+        fi
+    fi
+fi
+
+if [[ $install_skill -eq 1 ]]; then
+    # Render template: substitute @@RALPH_REPO@@ with the absolute repo path.
+    # Use | as sed delimiter since paths contain / — REPO shouldn't contain |.
+    sed "s|@@RALPH_REPO@@|$REPO|g" "$TEMPLATE" > "$SKILL_TARGET"
```
carbonscott: Nit: The `|` delimiter assumption (`REPO shouldn't contain |`) is *almost* always true on `~/codes/...` paths, but POSIX filenames allow any byte except `/` and NUL, so a path with a literal `|` would break the substitution silently (sed would parse it as the delimiter). Since you already documented the assumption, this is fine as-is. If you ever want to be bulletproof, the fix is a one-liner: escape the replacement before interpolating.

```bash
ESC_REPO=$(printf '%s\n' "$REPO" | sed 's/[|&\\]/\\&/g')
sed "s|@@RALPH_REPO@@|$ESC_REPO|g" "$TEMPLATE" > "$SKILL_TARGET"
```

Non-blocking — just flagging because you asked about it specifically.

install.sh
```diff
@@ -0,0 +1,151 @@
...
+#      with @@RALPH_REPO@@ substituted, so `/ralph-lnb` works in Claude Code chats.
+#
+# Idempotent: safe to re-run after moving or re-cloning the repo. Pass --force
+# to overwrite install artifacts that don't belong to ralph-wiggum-lnb.
+
+# --- Resolve this script's real location (symlink-safe) ---
+SOURCE="${BASH_SOURCE[0]}"
+while [ -L "$SOURCE" ]; do
+    DIR="$(cd "$(dirname "$SOURCE")" && pwd)"
+    SOURCE="$(readlink "$SOURCE")"
```
carbonscott, line 17: FYI: Plain `readlink "$SOURCE"` (no `-f`) is the right call here — it does single-level resolution on both BSD and GNU coreutils, and the while-loop walks levels manually. Portable and correct. No action needed; just confirming since you asked.

install.sh
```diff
@@ -0,0 +1,151 @@
...
+
+if [[ $install_skill -eq 1 ]]; then
+    # Render template: substitute @@RALPH_REPO@@ with the absolute repo path.
+    # Use | as sed delimiter since paths contain / — REPO shouldn't contain |.
+    sed "s|@@RALPH_REPO@@|$REPO|g" "$TEMPLATE" > "$SKILL_TARGET"
+    echo "Installed: $SKILL_TARGET (rendered with $REPO)"
+fi
+
+# --- PATH check (informational only) ---
+case ":$PATH:" in
```
carbonscott, line 127: FYI: `case ":$PATH:" in *":$BIN_DIR:"*)` is the canonical form and is correct. A `:` inside `$BIN_DIR` would break this, but a `:` in a PATH component is already unusable (the shell can't reference it), so the edge case is moot. Good as-is.

uninstall.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# --- Resolve this script's real location (symlink-safe) ---
+SOURCE="${BASH_SOURCE[0]}"
+while [ -L "$SOURCE" ]; do
+    DIR="$(cd "$(dirname "$SOURCE")" && pwd)"
+    SOURCE="$(readlink "$SOURCE")"
+    [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE"
+done
+REPO="$(cd "$(dirname "$SOURCE")" && pwd)"
+
```
carbonscott, line 15: Consider: `RALPH_BIN_DIR` asymmetry — if a user installs with `RALPH_BIN_DIR=/usr/local/bin ./install.sh` and later runs plain `./uninstall.sh` without re-exporting the var, uninstall checks `~/.local/bin/ralph` and silently leaves the real symlink orphaned. Options: (1) document in README that `RALPH_BIN_DIR` must be set at uninstall time too, (2) record the effective `BIN_DIR` in the skill file as a comment and read it back at uninstall, (3) accept the footgun as documented. Option (1) is probably the right call — it's just a README line. Non-blocking.

skill/SKILL.md.template
```diff
@@ -0,0 +1,129 @@
...
+
+The ralph repo this skill was generated from lives at:
+
+    @@RALPH_REPO@@
+
+If that path is stale (you moved or re-cloned the repo), re-run `./install.sh` from the new location to regenerate this file.
+
+## Parse arguments
+
+Parse `$ARGUMENTS` as a whitespace-separated list of `key value` pairs. Supported keys:
```
carbonscott, line 22: Nit: The `$ARGUMENTS` parsing instruction says "whitespace-separated list of `key value` pairs" and the table lists supported keys, but it doesn't say what to do with alternate forms a user might reasonably try: `max-iterations=5`, `--max-iterations 5`, or `-n 5`. Claude will probably do the sensible thing (treat them as the same thing) but pinning it down would make the skill more predictable. Could be as simple as: "Accept `key value`, `key=value`, or `--key value`. Unknown keys are ignored." Non-blocking.

carbonscott (commented):
## Review Summary

**Event:** COMMENT (GitHub blocks self-APPROVE; this would be APPROVE otherwise)

Clean, additive change. Two install scripts, one skill template, one README section — and zero edits to runtime scripts. The safety model is well-thought-out: symlink-identity check for the bin artifact, marker-string check for the rendered skill file. Both scripts walk `BASH_SOURCE` the same symlink-safe way the runners do. The PR description is one of the better ones I've read on this repo: explicit test plan, design rationale, out-of-scope called out.

The specific things you flagged me to look at:

### Things you asked about — verdicts

- **`sed "s|@@RALPH_REPO@@|$REPO|g"` safety.** POSIX filenames allow literally any byte except `/` and NUL, so a path containing `|` is legal on both macOS (APFS/HFS+) and Linux (ext/btrfs/xfs). In practice no one has a `|` in a `~/codes/...` path, and you documented the assumption inline, but it is a real edge case. Left a non-blocking nit with a one-liner fix.
- **`readlink "$SOURCE"` portability.** Correct and portable. Plain `readlink` (no `-f`) does single-level target resolution on both BSD and GNU coreutils — the while-loop walks levels manually. This is the canonical portable pattern. No issue.
- **Marker-based uninstall trade-off.** Reasonable. An HTML-comment marker is visible in the file; users who want to "adopt" it can delete the line; uninstall leaves adopted files alone. The alternative (checking for a substituted `@@RALPH_REPO@@` path) would be strictly more conservative but catches fewer legitimate custom edits. Your call is correct.
- **BSD-incompatible `sed` flags anywhere in the diff.** None. No `sed -i`. No GNU-only flags.
- **Safety-check asymmetry (symlink target vs marker string).** Fine. Different artifact types need different identity rules; a symlink's identity *is* its target path, a rendered file's identity is its content.
- **`--force` as a global flag.** Reasonable. Only two artifacts — per-artifact `--force-bin` / `--force-skill` would be over-engineered.
- **`install.sh --force --help` parsing.** Verified by trace. `for arg in "$@"` iterates in order. If `--force` comes first, FORCE=1 is set, then `--help` hits the help branch and `exit 0` terminates before any install action runs. If `--help` comes first, help prints and exits; FORCE never gets set but it doesn't matter. Behavior is: help always wins, nothing gets installed when `--help` is present. That matches your understanding and is defensible — help-wins is the least-surprise rule for CLI flags.
- **SKILL.md.template procedure sanity.** The procedure is clear and the `ALL_DONE`-before-`DONE` substring ordering is explicitly called out (load-bearing, and you nailed the reason). One minor ambiguity in the `$ARGUMENTS` parsing instructions — left as a nit.
- **README accuracy.** Matches the script behavior. One terminology nit left inline about calling the skill a "slash command".
- **PATH check via `case`.** `case ":$PATH:" in *":$BIN_DIR:"*)` is canonical. A `:` in `$BIN_DIR` would break it, but a `:` in a PATH directory is already unusable because the shell can't find it — so this is a non-issue, not even worth a nit.

### Other observations

- **`RALPH_BIN_DIR` asymmetry between install and uninstall.** If a user runs `RALPH_BIN_DIR=/usr/local/bin ./install.sh` and then later runs a plain `./uninstall.sh`, uninstall looks at `~/.local/bin/ralph` and silently misses the real target. Not a bug — both scripts read the env var — but it's an easy footgun. Worth a line in README ("re-export `RALPH_BIN_DIR` when uninstalling if you used it at install time") or a note in the install.sh final summary. Non-blocking.
- **Empty `$SKILL_DIR` on the unmanaged-skill path.** `mkdir -p "$SKILL_DIR"` runs before the safety check decides to skip; if the skill is unmanaged and `--force` isn't passed, you leave a freshly-created empty directory behind. Purely cosmetic. Optional.
- **`set -euo pipefail`**, symlink walk, PATH check, post-install summary, session-restart reminder — all good hygiene. Nothing I'd ask you to change.

### Tests

Your PR body lists 8 manual test scenarios run locally. The coverage is good: fresh install, idempotence, repo-move with and without `--force`, non-symlink refusal, uninstall, re-uninstall no-op, syntax check. Two test scenarios I'd suggest for follow-up (not blocking this PR):

1. Install with `RALPH_BIN_DIR=/tmp/custom-bin`, then uninstall with default env — verify the orphan scenario above and decide whether to fix in code or docs.
2. Fresh repo clone into a path containing a space (e.g. `~/my ralph/`). I don't see any unquoted `$REPO` in the diff, so it should work, but it's the kind of thing that bites you once in production.

Neither is blocking. Nice PR.

README.md
```diff
@@ -8,6 +8,34 @@ Based on the [Ralph Wiggum technique](https://ghuntley.com/ralph/) with
...
+```bash
+git clone https://github.com/carbonscott/ralph-wiggum-lnb ~/codes/ralph-wiggum-lnb
+~/codes/ralph-wiggum-lnb/install.sh
+```
+
+Two install artifacts:
+
+- `~/.local/bin/ralph` → symlink to the headless runner. Put `~/.local/bin`
+  on `$PATH` if it isn't already (`install.sh` warns you if not).
+- `~/.claude/skills/ralph-lnb/SKILL.md` → the `/ralph-lnb` slash command
```
carbonscott: Nit: This says "the `/ralph-lnb` slash command" but the PR description explicitly notes that skills replaced slash commands in January 2026 — the skill *auto-registers as* `/ralph-lnb`, which is subtly different. Users who read "slash command" might go looking for a `~/.claude/commands/` file. Suggest: "the `/ralph-lnb` skill trigger for Claude Code chats" or "the Claude Code skill that registers as `/ralph-lnb`". Non-blocking.

Issue #3, opened 2026-04-02
Switch task file from Markdown to JSON for agent safety


carbonscott:
Claude is less likely to inappropriately rewrite or mangle JSON files compared to Markdown. The rigid structure acts as a guardrail -- the only sanctioned mutation is flipping passes: false to true. Currently Lisa uses tasks.md (markdown with checkboxes). Switch to tasks.json following Ralph prd.json pattern.

PR #15, opened 2026-04-15
feat(runners): default --prompt to shared/PROMPT.md

### LINKED ISSUE #14: ralph runners require a local PROMPT.md copy for no good reason
## Problem

Both ralph runners hard-require `./PROMPT.md` in the project cwd and abort otherwise:

- `cc-headless/ralph.sh` — aborts at the prompt-file validation block if `./PROMPT.md` is missing.
- `cc/ralph-prep.sh` — same, used by the in-session flow.

The error messages both tell the user to `cp ~/codes/ralph-wiggum-lnb/shared/PROMPT.md .` before running. In practice that copy is pure ceremony: `shared/PROMPT.md` is the canonical template, ships with the checkout, and almost never needs per-project customization.

## Impact

After #13 shipped `install.sh` — which registers `ralph` / `/ralph-lnb` as ergonomic entry points — a freshly-installed `ralph` still fails on a clean project dir until the cp dance is done. The ergonomics gain of #13 is undermined: you can type `ralph` from anywhere, but you still have to pre-seed every project with a file copied from the repo.

Worse, the in-session driver (`/ralph-lnb`) has no way to pass a custom prompt at all — its SKILL definition only parses `max-iterations` and `task-file` from user arguments, so headless `ralph --prompt ./custom.md` has no equivalent.

## Proposal

1. Make `$PROMPT_FILE` default to `$SHARED_DIR/PROMPT.md` in both runners. `--prompt FILE` stays as an explicit override.
2. Add `--prompt` passthrough to the in-session drivers (`skill/SKILL.md.template` and `cc/RALPH-CC.md`) so users can type `/ralph-lnb max-iterations 5, prompt ./custom.md` and have the path forwarded to `ralph-prep.sh`.
3. Update all docs (root README, tests README, sandbox script, driver doc, skill template prereqs) to reflect the new workflow — most notably, drop every remaining `cp shared/PROMPT.md .` instruction.

## Backward-compat note

Projects that were silently relying on auto-pickup of a local `./PROMPT.md` will need to either delete the stale copy or pass `--prompt ./PROMPT.md` explicitly. The shared template is the source of truth going forward.

## Post-merge user action

Anyone who already ran `install.sh` must re-run it after pulling the merge — `install.sh` renders `skill/SKILL.md.template` into a **copy** at `~/.claude/skills/ralph-lnb/SKILL.md`, not a symlink, so the template edits don't propagate automatically. `install.sh` is idempotent, so it's a one-liner.

carbonscott:
## Summary

- Make `--prompt FILE` optional on both `ralph.sh` and `ralph-prep.sh`: when omitted, the runners auto-resolve `$SHARED_DIR/PROMPT.md` from the installed repo. The `cp shared/PROMPT.md .` ritual is gone.
- Add `prompt FILE` passthrough to the in-session drivers (`/ralph-lnb` and `cc/RALPH-CC.md`) so chat mode has the same escape hatch as headless mode.
- Purge every stale `cp shared/PROMPT.md .` instruction from the docs (root README, tests README, sandbox script, driver doc, skill template prereqs).

Closes #14.

## Behavior change

Projects that were silently relying on auto-pickup of a local `./PROMPT.md` will need to either delete the stale copy or pass `--prompt ./PROMPT.md` explicitly. The shared template is the source of truth going forward.

## Post-merge user action

`install.sh` renders `skill/SKILL.md.template` into a **copy** at `~/.claude/skills/ralph-lnb/SKILL.md`, not a symlink. Anyone who already installed must re-run `install.sh` after pulling to pick up the new `/ralph-lnb` behavior. `install.sh` is idempotent — the re-run is a one-liner.

## Test plan

- [x] `bash -n` on both modified runners
- [x] `ralph --help` shows `(default: repo's shared/PROMPT.md)`
- [x] `ralph-prep.sh --help` shows the same
- [x] `ralph-prep.sh` from a sandbox with only `tasks.json` (no local `PROMPT.md`) emits the filled prompt on stdout, matching `shared/PROMPT.md` after FILL substitution
- [x] `./install.sh` re-renders `~/.claude/skills/ralph-lnb/SKILL.md` with `@@RALPH_REPO@@` substituted; eyeballed the output — prompt row, prereq #3, and Bash() call all correct
- [x] **Headless end-to-end:** `ralph --max-iterations 3` in a sandbox containing only `tasks.json` — runs to `ALL_DONE`, produces `hello.txt` with both expected lines, flips both stories to `passes: true`, populates `.lnb/`. Confirmed against the smoke-test fixture.
- [x] **In-session end-to-end:** `/ralph-lnb max-iterations 3` in a sandbox containing only `tasks.json` — same expectations, same result. Confirms the new `prompt=P` announce line and the default-template auto-resolution work in chat mode.
- [ ] In-session custom prompt path: `/ralph-lnb max-iterations 1, prompt ./custom.md` forwards `--prompt` to `ralph-prep.sh` and runs (manual follow-up)
- [ ] In-session missing prompt path: `/ralph-lnb max-iterations 1, prompt ./missing.md` surfaces `ralph-prep.sh`'s "Prompt file not found" error and stops (manual follow-up)

Unchecked items are the two `prompt=` override scenarios — the happy-path default behavior is confirmed in both modes, so the remaining checks are for the escape-hatch branch only and can be validated post-merge without blocking.

## Review follow-up (commit 2, `3df3267`)

Addressed the two substantive nits from the subagent review:

- **README migration note:** new "Upgrading from an older checkout?" sentence in Quick Start warns existing users about stale `./PROMPT.md`.
- **Announce-line symmetry:** `cc/RALPH-CC.md` and `skill/SKILL.md.template` Setup step 3 now include `prompt=P` in the `Starting ralph loop…` announcement, matching the headless runner's `Prompt:` banner line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,17 +35,22 @@ location with `RALPH_BIN_DIR=/usr/local/bin ./install.sh`. Undo with
 `./uninstall.sh` — if you overrode `RALPH_BIN_DIR` on install, pass
 the same value on uninstall so it can find the symlink to remove.
 
-If you prefer not to install, the absolute-path invocations in the
-Quick Start below still work.
+If you prefer not to install, you can still run the scripts by
+absolute path: `~/codes/ralph-wiggum-lnb/cc-headless/ralph.sh` for
+headless mode, or `follow ~/codes/ralph-wiggum-lnb/cc/RALPH-CC.md` in
+a Claude Code session.
 
 ## Quick Start
 
 ```bash
 # In your project directory:
-cp ~/codes/ralph-wiggum-lnb/shared/PROMPT.md .
 cp ~/codes/ralph-wiggum-lnb/shared/tasks.json.example tasks.json
 # Edit tasks.json with your stories
 ```
+
+The runner uses `shared/PROMPT.md` from the installed repo by default.
+Copy it locally and pass `--prompt ./PROMPT.md` only if you want to
+customize the template.
 
 The runner auto-initializes `.lnb/` with the coding schema on the first
 iteration — no manual `lab-notebook init` needed.
@@ -55,19 +60,22 @@ Then pick one runner:
 **Headless** (uses `claude -p`):
 
 ```bash
-~/codes/ralph-wiggum-lnb/cc-headless/ralph.sh --max-iterations 5
+ralph --max-iterations 5
 ```
 
 **Inside a Claude Code session** (uses the `Agent()` subagent tool — no `-p` needed):
 
 Start Claude Code in `acceptEdits` mode, then in the session:
 
-> follow ~/codes/ralph-wiggum-lnb/cc/RALPH-CC.md, max-iterations 5
+> /ralph-lnb max-iterations 5
 
-No extra files need to live in your project dir — `PROMPT.md` +
-`tasks.json` is the entire footprint. Helper scripts, the shared lib,
-and the notebook schema all stay in the repo and are invoked or sourced
-by absolute path.
+(Restart any running Claude Code sessions after install — skills load
+at session start.)
+
+`tasks.json` is the entire footprint in your project dir. The prompt
+template, shared lib, helper scripts, and notebook schema all stay in
+the repo and are invoked or sourced by `ralph` / `/ralph-lnb` (or by
+absolute path if you skipped install).
 
 ## How It Works
 
@@ -179,20 +187,21 @@ LAB_NOTEBOOK_DIR=.lnb lab-notebook sql \
   "SELECT content FROM entries WHERE type='pattern' ORDER BY ts"
 ```
 
-## `cc-headless/ralph.sh` options
+## Headless runner options (ralph)
 
 ```
 --max-iterations N      Safety cap (default: 10)
---prompt FILE           Prompt template (default: PROMPT.md)
+--prompt FILE           Custom prompt template (default: repo's shared/PROMPT.md)
 --task-file FILE        Task file (default: tasks.json)
 --notebook DIR          Notebook directory (default: .lnb)
 --context SLUG          Notebook context (default: from branch)
 --archive-dir DIR       Archive directory (default: archive/)
 ```
 
-For the Claude Code session runner, `max-iterations`, `task-file`, and
-related parameters are passed inline when invoking the driver doc — see
-`cc/RALPH-CC.md`.
+For the Claude Code session runner, invoke it as `/ralph-lnb
+max-iterations N` (with `task-file` and other parameters passed
+inline). See `cc/RALPH-CC.md` for the underlying driver doc and
+stop-condition semantics.
 
 ## Archive
 
--- a/README.md
+++ b/README.md
@@ -50,7 +50,10 @@ cp ~/codes/ralph-wiggum-lnb/shared/tasks.json.example tasks.json
 
 The runner uses `shared/PROMPT.md` from the installed repo by default.
 Copy it locally and pass `--prompt ./PROMPT.md` only if you want to
-customize the template.
+customize the template. **Upgrading from an older checkout?** A stale
+`./PROMPT.md` in your project dir is no longer auto-picked up — delete
+it if you never customized it, or pass `--prompt ./PROMPT.md`
+explicitly if you did.
 
 The runner auto-initializes `.lnb/` with the coding schema on the first
 iteration — no manual `lab-notebook init` needed.
--- a/cc-headless/ralph.sh
+++ b/cc-headless/ralph.sh
@@ -3,7 +3,7 @@ set -euo pipefail
 
 # --- Defaults ---
 MAX_ITERATIONS=10
-PROMPT_FILE="PROMPT.md"
+PROMPT_FILE=""
 TASK_FILE="tasks.json"
 NOTEBOOK_DIR=".lnb"
 CONTEXT=""
@@ -32,7 +32,7 @@ file (tasks.json) and a lab-notebook.
 
 Options:
   --max-iterations N      Safety cap (default: 10)
-  --prompt FILE           Prompt template (default: PROMPT.md)
+  --prompt FILE           Custom prompt template (default: repo's shared/PROMPT.md)
   --task-file FILE        Task file with stories (default: tasks.json)
   --notebook DIR          Lab-notebook directory (default: .lnb)
   --context SLUG          Notebook context (default: derived from branch)
@@ -66,10 +66,14 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 
+# Default to the repo's shared template if --prompt wasn't passed.
+if [[ -z "$PROMPT_FILE" ]]; then
+    PROMPT_FILE="$SHARED_DIR/PROMPT.md"
+fi
+
 # --- Validate ---
 if [[ ! -f "$PROMPT_FILE" ]]; then
     echo "Error: Prompt file '$PROMPT_FILE' not found." >&2
-    echo "Copy the template: cp $SHARED_DIR/PROMPT.md ." >&2
     exit 1
 fi
 
@@ -134,6 +138,7 @@ echo "Project:    ${PROJECT:-<none>}"
 echo "Branch:     ${BRANCH:-<none>}"
 echo "Context:    $CONTEXT"
 echo "Task file:  $TASK_FILE"
+echo "Prompt:     $PROMPT_FILE"
 echo "Notebook:   $NOTEBOOK_DIR"
 echo "Max iter:   $MAX_ITERATIONS"
 echo "============="
--- a/cc/RALPH-CC.md
+++ b/cc/RALPH-CC.md
@@ -5,23 +5,29 @@ designed to run entirely inside a live Claude Code chat session using
 the `Agent()` subagent tool instead of headless `claude -p`. Use this
 when `-p` mode is unavailable or restricted.
 
-Same `PROMPT.md`, same `tasks.json`, same `.lnb/` notebook, same
-`archive/` directory, same per-iteration semantics. The only thing that
-changes is who spawns the per-iteration agent.
+Same prompt template, same `tasks.json`, same `.lnb/` notebook, same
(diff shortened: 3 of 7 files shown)


cc-headless/ralph.sh
```diff
@@ -134,6 +138,7 @@ echo "Project:    ${PROJECT:-<none>}"
 echo "Branch:     ${BRANCH:-<none>}"
 echo "Context:    $CONTEXT"
 echo "Task file:  $TASK_FILE"
+echo "Prompt:     $PROMPT_FILE"
```
carbonscott, line 141: FYI: Nice addition. Now that the default prompt path is no longer `./PROMPT.md`, having the resolved path in the banner means users can see at a glance whether they're running the shared template or an override. Small change, high signal.

README.md
```diff
@@ -35,18 +35,23 @@ location with `RALPH_BIN_DIR=/usr/local/bin ./install.sh`. Undo with
...
 
 ```bash
 # In your project directory:
-cp ~/codes/ralph-wiggum-lnb/shared/PROMPT.md .
 cp ~/codes/ralph-wiggum-lnb/shared/tasks.json.example tasks.json
 # Edit tasks.json with your stories
 ```
 
+The runner uses `shared/PROMPT.md` from the installed repo by default.
+Copy it locally and pass `--prompt ./PROMPT.md` only if you want to
```
carbonscott, line 52: Nit: This paragraph tells *new* users exactly what they need to know (copy locally only if you want to customize). But readers migrating from the old behavior — who already have a `./PROMPT.md` sitting in their project — will not realize from this paragraph that the runner now *ignores* that file unless they pass `--prompt ./PROMPT.md` explicitly. The commit message and PR body cover this, but README-only readers will miss it.

Consider a one-liner like: `Projects that used to keep a local ./PROMPT.md should delete it or pass --prompt ./PROMPT.md explicitly — the runner no longer auto-picks it up.` Not blocking; release notes can carry this.

cc/RALPH-CC.md
```diff
@@ -47,8 +52,10 @@ minimum specified.
 
 ### Setup
 
-1. Parse `max-iterations` and `task-file` from the user's message.
-   Apply defaults for anything unspecified.
+1. Parse `max-iterations`, `task-file`, and `prompt` from the user's
+   message. Apply defaults for anything unspecified. `prompt` is
+   optional — if the user did not supply it, leave it unset and omit
+   `--prompt` from the `ralph-prep.sh` call in Loop step 1.
```
carbonscott, line 58: Consider: Setup step 1 now parses `prompt`, but step 3 — the announce line just below, at line 68 of the rendered file, not in this diff — still only reports `max-iterations`, `task-file`, and `context`. The headless runner gained a `Prompt:` banner line in this same PR; consider mirroring that here so the main session surfaces which template it's actually using.

Something like: `"Starting ralph loop, max-iterations=N, task-file=X, prompt=<prompt or default>, context=<context>"`.

Not blocking — the default is well-known and the override is rare — but it would close a small observability gap.

skill/SKILL.md.template
```diff
@@ -52,7 +52,7 @@ When invoked, the main session (you) must do the following. Do not deviate, do n
 
 ### Setup
 
-1. Parse `max-iterations` and `task-file` from `$ARGUMENTS`. Apply the defaults above for anything unspecified.
+1. Parse `max-iterations`, `task-file`, and `prompt` from `$ARGUMENTS`. Apply the defaults above for anything unspecified. `prompt` is optional — if the user did not supply it, leave the variable unset and omit `--prompt` from the `ralph-prep.sh` call in Loop step 1.
```
carbonscott, line 55: Nit: Same observation as on `cc/RALPH-CC.md` Setup step 1 — the parse now includes `prompt`, but the announce line in Setup step 3 (line 63 of this template, not in the diff) still only mentions `max-iterations`, `task-file`, `context`. Consider adding `prompt` to the announce string so users can verify which template is being used.

If you change one driver, change both so they stay in lockstep — this skill template and `cc/RALPH-CC.md` are deliberately symmetric.

skill/SKILL.md.template
```diff
@@ -69,9 +69,11 @@ For `i` in `1..max-iterations`:
 1. **Build the prompt.** Call:
 
    ```
-   Bash("@@RALPH_REPO@@/cc/ralph-prep.sh --iteration i --max-iterations N --task-file <task-file>")
+   Bash("@@RALPH_REPO@@/cc/ralph-prep.sh --iteration i --max-iterations N --task-file <task-file>[ --prompt <prompt>]")
```
carbonscott, line 72: FYI: The conditional syntax `[ --prompt <prompt>]` inside the `Bash(...)` string, combined with the plain-English explanation on the next paragraph ("Append `--prompt <prompt>` only if the user supplied one..."), reads clearly to a human and should also be unambiguous to the main-session model. Good. The fact that you repeat the same pattern verbatim in `cc/RALPH-CC.md` is the right move — the two drivers need to drift in sync, not independently.

No change requested.

cc/RALPH-CC.md
```diff
@@ -5,23 +5,29 @@ designed to run entirely inside a live Claude Code chat session using
...
+If you skipped `install.sh`, the long form still works:
+
+> follow ~/codes/ralph-wiggum-lnb/cc/RALPH-CC.md, max-iterations 10
 
 Defaults when unspecified:
 
 - `max-iterations` = 10
 - `task-file` = `tasks.json`
-- `prompt` = `PROMPT.md`
+- `prompt` = the installed repo's `shared/PROMPT.md`
```
carbonscott, line 30: Nit: The defaults list now says `prompt` = `the installed repo's shared/PROMPT.md`, which is accurate for installed users but slightly opaque for someone who invoked via the long-form `follow ~/codes/ralph-wiggum-lnb/cc/RALPH-CC.md` fallback shown immediately above. For them it's the *cloned* repo, not the *installed* repo (install.sh renders a copy of the skill, not a symlink to this file — so the long-form invocation bypasses install entirely). Consider: `prompt = the ralph-wiggum-lnb repo's shared/PROMPT.md`.

Minor — the two are the same file 99% of the time — but the wording could be less install-centric in a doc that explicitly supports the no-install path.

carbonscott (commented):
## Review Summary

**Overall:** Clean, well-scoped change that removes a pointless ceremony (`cp shared/PROMPT.md .`) and tightens the ergonomic surface across all four entry points (headless script, chat driver, skill template, tests). Docs are in lockstep with the behavior change, the commit message is excellent, and the PR description nails both the migration story and the post-merge install action.

*(Posting as COMMENT rather than APPROVE because GitHub blocks self-approval — this would have been an APPROVE otherwise. All findings are Nit / Consider / FYI; nothing blocking.)*

### Good things

- Commit message is textbook: *what*, *why*, per-file summary, explicit behavior-change callout, explicit post-merge user action. This is the quality to see more of.
- One commit for one logical unit across 7 files is the right call here — splitting the runner change from the doc alignment would have left the tree in an inconsistent state between commits.
- The new `Prompt:` line in the headless banner (`cc-headless/ralph.sh:141`) is a small but high-value addition — now the resolved prompt path is visible at a glance, which matters more precisely *because* the default is no longer `./PROMPT.md`.
- The in-session `prompt` passthrough is implemented symmetrically in both `skill/SKILL.md.template` and `cc/RALPH-CC.md`, with the same conditional `[ --prompt <prompt>]` syntax followed by the same explanatory sentence. The disambiguation is unambiguous enough that the main-session model should handle it correctly.
- `tests/setup-sandbox.sh` net -1 line, and `tests/README.md` gotchas bullet #2 was updated in the same PR so the smoke-test docs stay truthful. No orphaned instructions.

### Key findings

**Documentation (nit-level):**

- The README still has two places where `PROMPT.md` is mentioned as a top-level project-dir artifact: the ASCII diagram around line 83 (`tasks.json + .lnb/ + PROMPT.md`) and the "Two runners, same loop" blurb around line 103 (`share the same PROMPT.md, tasks.json`). Neither is in this PR's diff, so no inline comment — but after this PR, `PROMPT.md` no longer lives in the project dir by default. These now read slightly out of sync with the new mental model. Not blocking; worth a follow-up touch-up pass.
- The README Quick Start tells *new* users about the new default, but doesn't warn *existing* users that "your local `./PROMPT.md` will no longer be auto-picked up." The commit message and PR body cover this, but README-only readers will miss it. See the inline comment on `README.md:52`.

**Consistency (nit-level):**

- The chat-mode driver's Setup step 1 now parses `prompt`, but the Setup step 3 announce string in both `cc/RALPH-CC.md` (around line 68) and `skill/SKILL.md.template` (around line 63) still only prints `max-iterations=N, task-file=X, context=<context>`. The headless runner banners the prompt path in this same PR — the in-session drivers now have the info but don't surface it. Note: those lines are not in the diff, so the inline comments below are attached to the nearest in-diff lines (the parse step) with a pointer forward. See inline comments on `cc/RALPH-CC.md:58` and `skill/SKILL.md.template:55`.

**No issues:**

- Grep confirms zero remaining `cp shared/PROMPT.md .` strings. Author's greps were clean; fresh pair of eyes also clean.
- Sandbox flow: `tests/README.md` still describes the smoke test correctly — the runners now resolve `shared/PROMPT.md` themselves, so the flow still works end-to-end. `tests/setup-sandbox.sh` no longer creates the symlink, and the gotchas bullet was rewritten in the same PR to match.
- Conditional Bash() syntax (`[ --prompt <prompt>]`): the square brackets are a common "optional arg" convention, and the explanatory sentence right after removes any ambiguity. The main-session model will read both and do the right thing.
- `install.sh` re-run messaging: the PR body calls it out; that's the right channel. Release notes > README for one-time migration steps.

### PR description feedback

The description is one of the better ones. First line summarizes the change, the Behavior change section explicitly flags the backward-incompat, the Post-merge user action section tells existing installs exactly what to do, and the test plan distinguishes what was verified from what is deferred to manual smoke tests post-merge with a one-sentence justification. No changes needed.

6 inline comments below — all Nit / Consider / FYI. None blocking.

PR #17, opened 2026-04-17
docs: use $RALPH_REPO for clone-path references

carbonscott:
## Summary
- Replace hardcoded `~/codes/ralph-wiggum-lnb` with `$RALPH_REPO` in all user-facing snippets so they copy-paste verbatim regardless of clone location.
- README Install section grows an `export RALPH_REPO=...` preamble + persistence note; skip-install bullet and Quick Start `cp` use the var.
- `cc/RALPH-CC.md` gains a new Prereq item 1 requiring `RALPH_REPO`; the runtime `Bash("$RALPH_REPO/cc/ralph-prep.sh …")` call in Loop step 1 drops the old `$HOME`-vs-tilde workaround.
- `install.sh` `--help` heredoc mirrors the new convention; post-install prints a concrete `export RALPH_REPO="<abs-path>"` tip for users who want to add it to their rc.
- No runtime code touched. `skill/SKILL.md.template` already parameterizes via `@@RALPH_REPO@@` substitution, so installed users continue to see concrete absolute paths.

## Test plan
- [x] `install.sh --help` renders literal `$RALPH_REPO` (single-quoted heredoc).
- [x] Re-running `install.sh` prints the tip with `$REPO` expanded to the concrete absolute path.
- [x] `grep -rn '~/codes/ralph-wiggum-lnb\|/ralph-wiggum-lnb/'` returns only the intentional example-value assignments in the three setup snippets.
- [ ] Reviewer sanity-check: docs read naturally end-to-end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,17 @@ Verify with `lab-notebook --help`. Update later with
 ## Install
 
 ```bash
-git clone https://github.com/carbonscott/ralph-wiggum-lnb ~/codes/ralph-wiggum-lnb
-~/codes/ralph-wiggum-lnb/install.sh
+# Pick any path; snippets below reference $RALPH_REPO so they copy-paste verbatim.
+export RALPH_REPO=~/codes/ralph-wiggum-lnb
+
+git clone https://github.com/carbonscott/ralph-wiggum-lnb "$RALPH_REPO"
+"$RALPH_REPO/install.sh"
 ```
+
+Add the `export` line to your `~/.zshrc` or `~/.bashrc` if you want it
+to persist across new terminals. The installed `ralph` binary and
+`/ralph-lnb` skill don't need it — only snippets in this README and
+`cc/RALPH-CC.md` do.
 
 Two install artifacts:
 
@@ -55,15 +63,14 @@ location with `RALPH_BIN_DIR=/usr/local/bin ./install.sh`. Undo with
 the same value on uninstall so it can find the symlink to remove.
 
 If you prefer not to install, you can still run the scripts by
-absolute path: `~/codes/ralph-wiggum-lnb/cc-headless/ralph.sh` for
-headless mode, or `follow ~/codes/ralph-wiggum-lnb/cc/RALPH-CC.md` in
-a Claude Code session.
+absolute path: `"$RALPH_REPO/cc-headless/ralph.sh"` for headless mode,
+or `follow $RALPH_REPO/cc/RALPH-CC.md` in a Claude Code session.
 
 ## Quick Start
 
 ```bash
 # In your project directory:
-cp ~/codes/ralph-wiggum-lnb/shared/tasks.json.example tasks.json
+cp "$RALPH_REPO/shared/tasks.json.example" tasks.json
 # Edit tasks.json with your stories
 ```
 
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ location with `RALPH_BIN_DIR=/usr/local/bin ./install.sh`. Undo with
 the same value on uninstall so it can find the symlink to remove.
 
 If you prefer not to install, you can still run the scripts by
-absolute path: `"$RALPH_REPO/cc-headless/ralph.sh"` for headless mode,
+absolute path: `$RALPH_REPO/cc-headless/ralph.sh` for headless mode,
 or `follow $RALPH_REPO/cc/RALPH-CC.md` in a Claude Code session.
 
 ## Quick Start
--- a/cc/RALPH-CC.md
+++ b/cc/RALPH-CC.md
@@ -21,7 +21,7 @@ In a Claude Code chat, type one of:
 
 If you skipped `install.sh`, the long form still works:
 
-> follow ~/codes/ralph-wiggum-lnb/cc/RALPH-CC.md, max-iterations 10
+> follow $RALPH_REPO/cc/RALPH-CC.md, max-iterations 10
 
 Defaults when unspecified:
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -55,15 +63,14 @@ location with `RALPH_BIN_DIR=/usr/local/bin ./install.sh`. Undo with
 the same value on uninstall so it can find the symlink to remove.
 
 If you prefer not to install, you can still run the scripts by
-absolute path: `~/codes/ralph-wiggum-lnb/cc-headless/ralph.sh` for
-headless mode, or `follow ~/codes/ralph-wiggum-lnb/cc/RALPH-CC.md` in
-a Claude Code session.
+absolute path: `"$RALPH_REPO/cc-headless/ralph.sh"` for headless mode,
```
carbonscott: Nit: quoting is inconsistent within this sentence — `"$RALPH_REPO/cc-headless/ralph.sh"` is quoted but `$RALPH_REPO/cc/RALPH-CC.md` on the next line isn't. Both are fine in isolation (the first is a bash snippet, the second is a "follow ..." instruction to Claude Code), but a reader scanning quickly may wonder if the difference is meaningful. Consider either both quoted or both bare for visual consistency.

cc/RALPH-CC.md
```diff
@@ -21,7 +21,7 @@ In a Claude Code chat, type one of:
 
 If you skipped `install.sh`, the long form still works:
 
-> follow ~/codes/ralph-wiggum-lnb/cc/RALPH-CC.md, max-iterations 10
+> follow $RALPH_REPO/cc/RALPH-CC.md, max-iterations 10
```
carbonscott, line 24: FYI: this line is what the user types into a Claude Code chat, not a shell. Claude Code generally does expand `$VAR` when constructing `Bash()` calls, so this works in practice — but a reader who's never used the long form may briefly wonder whether they should pre-expand it themselves. The new Prereq item 1 makes the intent clear further down, so this is fine as-is.

cc/RALPH-CC.md
```diff
@@ -32,17 +32,21 @@ Defaults when unspecified:
 
 ## Prerequisites
 
-1. The main Claude Code session must be in `acceptEdits` permission mode
+1. `RALPH_REPO` must be exported in the shell where Claude Code runs
```
carbonscott, line 35: Optional: consider whether the Setup section should include a runtime guard step — e.g. an early `Bash("test -n \"$RALPH_REPO\" || (echo 'RALPH_REPO is not exported; see Prereq item 1' >&2; exit 1)")` before the notebook-context derivation. Upside: fail-fast with an actionable message rather than letting iteration 1's `ralph-prep.sh` call fail with an empty-path error. Downside: adds a Setup step, and this Prereq item plus the updated prose in Loop step 1 already spell it out clearly. Happy either way — the current prose-based approach is defensible for a doc-first workflow.

carbonscott (commented):
## Review Summary

**Overall:** Clean, focused docs refactor. Replacing hardcoded `~/codes/ralph-wiggum-lnb` with `$RALPH_REPO` makes the snippets genuinely copy-paste portable, and scoping the env var to user-facing docs (not runtime scripts or the installed skill) is the right call. Event is COMMENT rather than APPROVE because GitHub blocks self-approval; treat this review as an LGTM with non-blocking suggestions.

### Key Findings

**Design / scope**
- Good call leaving `shared/ralph-lib.sh`, `cc-headless/ralph.sh`, `cc/ralph-prep.sh`, and `skill/SKILL.md.template` untouched. Installed users get concrete absolute paths via `@@RALPH_REPO@@` substitution; only doc readers need to set the var. That's the right boundary.
- Printing the concrete `export RALPH_REPO="$REPO"` tip post-install (unquoted heredoc so `$REPO` expands) is a nice touch — users get a ready-to-paste line with their actual absolute path, not a literal tilde.

**Completeness**
- `grep -n '~/codes/ralph-wiggum-lnb\|/ralph-wiggum-lnb/'` on the PR branch returns exactly three hits: `README.md:34`, `cc/RALPH-CC.md:36`, and `install.sh:33` — all intentional example assignments. Nothing missed.

**UX clarity**
- First-time reader path works: the README Install section opens with the export, and the follow-up paragraph explicitly tells them the installed `ralph` and `/ralph-lnb` don't need it. The RALPH-CC.md Prereq item 1 is explicit about the same requirement for CC sessions.

**Runtime concern (the `Bash("$RALPH_REPO/...")` call in RALPH-CC.md Loop step 1)**
- The author's updated prose addresses this: "an unset variable yields an empty path and the call fails with a clear error." Combined with Prereq item 1, this is sufficient for the common case. A Setup-step runtime guard would be more defensive but adds a step for marginal benefit — see the optional inline comment on the new Prereq item.

### PR Description Feedback
Description is excellent — clear scope, rationale for the design split (`@@RALPH_REPO@@` substitution vs. `$RALPH_REPO` env var), and a verifiable test plan with grep assertion.

4 inline comments left (all Nit/Optional/FYI — none blocking).

install.sh
```diff
@@ -138,6 +139,10 @@ Done. Quick usage:
   ralph --max-iterations 3                 # headless mode
   /ralph-lnb max-iterations 3              # in a Claude Code chat
 
+Tip: if you plan to use RALPH-CC.md or the uninstalled fallback, add
+  export RALPH_REPO="$REPO"
```
carbonscott, line 143: Nice — printing the concrete resolved `$REPO` path here (rather than a literal `~/codes/ralph-wiggum-lnb`) means the user can copy the line into their rc file verbatim regardless of where they cloned. The unquoted heredoc (`cat <<EOF`) makes this work correctly, and it pairs well with the README note about persisting the export.

PR #18, opened 2026-04-17
refactor: collapse chat entry to /ralph-lnb skill, align prompt build

carbonscott:
## Summary

Two tightly-related changes:

### A. Collapse the chat entry point
- **Delete `cc/RALPH-CC.md`.** It was a parallel "follow this file" fallback for users who skipped `install.sh`. The skill template (`skill/SKILL.md.template`) is now the single source of truth; `install.sh` renders it with concrete absolute paths. PR #17 tried to keep RALPH-CC.md portable with `$RALPH_REPO` — stepping back, the whole fallback is not worth a drifting duplicate doc + env var + extra test surface.
- **Revert all `$RALPH_REPO` references** from README, `install.sh` `--help` + post-install summary, and `ralph-prep.sh` comments. Reverts #17's additions.
- **Rewrite README's "Two runners, same loop"** section to reference `/ralph-lnb` directly. Migrate the "Differences from `ralph.sh`" comparison table from RALPH-CC.md (useful decision aid — worth preserving).
- Update `cc/ralph-prep.sh`, `tests/setup-sandbox.sh`, `tests/README.md` to drop RALPH-CC.md refs.
- Update Layout section to reflect 4 dirs (was 3) now that `skill/` is acknowledged.

Headless (`cc-headless/ralph.sh` invoked by absolute path) remains a no-install option. The chat runner now requires `install.sh`.

### B. Align chat-mode prompt build with headless
Both runners call `build_prompt` from `shared/ralph-lib.sh` against the same template — prompt *content* was identical modulo one subtle ordering bug:
- `cc-headless/ralph.sh:151-154` queries history then builds prompt then logs `start`.
- `cc/ralph-prep.sh` previously logged `start` then queried history then built prompt.

So on iteration N>=2, chat mode's `recent_history` block included its own iteration-start entry (pushing out one older row); headless's didn't. Reorder `ralph-prep.sh` to match headless — now both runners build byte-identical prompts given the same notebook state. Log prefix unified to `ralph-lnb:` at the same time.

## Test plan
- [x] Grep across the repo: no user-facing `RALPH-CC.md`, `$RALPH_REPO`, or `ralph-cc:` hits (only `@@RALPH_REPO@@` substitution tokens in the template + install.sh comments).
- [x] `./install.sh` succeeds; `./install.sh --help` shows clean output (no RALPH_REPO mentions); post-install summary drops the tip block.
- [x] `./install.sh` renders `~/.claude/skills/ralph-lnb/SKILL.md` correctly (template unchanged, output byte-identical).
- [ ] Reviewer sanity: README's rewritten "Two runners" section reads naturally end-to-end.
- [ ] Prompt-parity spot-check: with a prepared `.lnb/`, run `cc/ralph-prep.sh --iteration 2` and verify `recent_history` matches what headless would produce for the same iteration.

Generated with Claude Code


cc/ralph-prep.sh
```diff
@@ -108,12 +108,15 @@ fi
...
-if [[ -n "$MAX_ITERATIONS" ]]; then
-    log_to_notebook "start" "ralph-cc: starting iteration $ITERATION/$MAX_ITERATIONS"
-else
-    log_to_notebook "start" "ralph-cc: starting iteration $ITERATION"
-fi
-
 # --- Emit filled prompt to stdout ---
+# Query history BEFORE logging "start" so the recent_history block sees
+# only prior iterations' entries — matches cc-headless/ralph.sh ordering
+# (see ralph.sh:151-154) so both runners build byte-identical prompts.
```
carbonscott: Nice comment — explains *why* the ordering matters and pins it to the exact `ralph.sh` line range. This is exactly the kind of comment that pays off six months later when someone considers moving the `log_to_notebook` call back up for tidiness and needs to know it's load-bearing.

Nit: the reference `ralph.sh:151-154` is slightly off — the matching lines in `cc-headless/ralph.sh` are 150-154 (the comment `# Harness does RECALL: query notebook and inject into prompt` sits at 150, then `history=`, `prompt=`, blank, `log_to_notebook`). Not worth a follow-up commit on its own, but mention if you re-touch this file.

README.md
```diff
@@ -229,8 +234,8 @@ LAB_NOTEBOOK_DIR=.lnb lab-notebook sql \
 
 For the Claude Code session runner, invoke it as `/ralph-lnb
 max-iterations N` (with `task-file` and other parameters passed
-inline). See `cc/RALPH-CC.md` for the underlying driver doc and
-stop-condition semantics.
+inline). See `skill/SKILL.md.template` for the underlying driver doc
```
carbonscott: Nit: pointing readers at `skill/SKILL.md.template` for stop-condition semantics is technically correct, but a new reader probably wants the *rendered* skill they actually see in Claude Code (`~/.claude/skills/ralph-lnb/SKILL.md`) — the template has `@@RALPH_REPO@@` placeholders that look noisy when read standalone. Consider pointing at the rendered path, or just say "See the `/ralph-lnb` skill" and let readers find it. Non-blocking.

carbonscott (commented):
## Review Summary

**Overall:** Would approve — this is a clean refactor with a legitimate bug fix attached. Switching to `event: COMMENT` only because GitHub forbids approving your own PR. Clean refactor. Collapsing the chat entry point to a single source of truth (the rendered skill) removes real drift risk, and the `ralph-prep.sh` reorder is a correct fix for a subtle history-injection bug.

### Key Findings

**Design / Functionality**
- The reorder in `cc/ralph-prep.sh` (archive -> ensure_notebook -> query_recent_history -> build_prompt -> log_to_notebook "start") correctly mirrors `cc-headless/ralph.sh:150-154`. Both runners now compute `history` before the iteration's own `start` entry is written, so `recent_history` contains the same rows on iteration N>=2. The diagnosis in the PR body (one-row drift where chat mode's block included its own iteration-start entry) matches what the old code actually did.
- Side-effect ordering is equivalent to headless: in `ralph.sh` the `start` log fires before `claude -p` spawns; in `ralph-prep.sh` it fires after `build_prompt` writes to stdout but before the caller reads stdout back and hands it to `Agent()`. Since `log_to_notebook` writes to the SQLite DB (not stdout) and the script exits cleanly, the Agent() call still sees the start row logged first — functionally identical to headless.
- Deletion of `cc/RALPH-CC.md` is the right call. A "follow this file" doc that duplicates the skill template was always going to drift; PR #17's `$RALPH_REPO` patch made that drift more expensive to maintain rather than solving it.

**Completeness of content migration**
- The "Differences from `ralph.sh`" table landed in README at lines 139-147 verbatim — confirmed.
- The "Relationship to `shared/ralph-lib.sh`" paragraph from RALPH-CC.md is covered by (a) README's Files table entry for `shared/ralph-lib.sh` ("Shared bash helpers sourced by both runners") and (b) SKILL.md.template's "Relationship to the headless runner" section (line 127-129), which explicitly states "A change to `shared/ralph-lib.sh` affects both runners." No material loss.
- Grep confirmed independently: no user-facing `RALPH-CC.md`, `$RALPH_REPO`, or `ralph-cc:` hits remain. Only matches are `install.sh` comments (lines 8, 117, 122) and `@@RALPH_REPO@@` substitution tokens in `skill/SKILL.md.template`. Clean.

**README flow**
- The rewritten "Two runners, same loop" section reads naturally: two bullets name the entry points, the table gives the detailed comparison, and the closing sentence points at both state stores. The Layout section's jump from 3 to 4 directories is a good call — `skill/` was already present and pretending otherwise was misleading.
- One minor: line 235-238 points users at `skill/SKILL.md.template` for stop-condition semantics. Technically correct (the procedure does live there), but most readers will read the *rendered* `~/.claude/skills/ralph-lnb/SKILL.md`. Nit-only.

**Breakage**
- Old `follow $RALPH_REPO/cc/RALPH-CC.md` invocation no longer works — `tests/setup-sandbox.sh` correctly drops its advertisement. No other callers found.
- `uninstall.sh` was not touched; if a user installed before this PR and uninstalls after, the existing uninstall logic should still find the symlink and the skill dir by path. Not verified in this review but out of scope for this PR.

### PR Description Feedback
Description is excellent — first line summarizes the refactor, body splits the two logical changes (A and B) with rationale, cites the exact line range being mirrored, and notes which bug the reorder fixes. Test plan has concrete checks. Model PR description.

2 inline comments left (both Nit). Self-merge when ready.

carbonscott:
## Smoke test results

Ran both modes in fresh sandboxes after a clean `./uninstall.sh && ./install.sh` on this branch.

- **Headless sandbox**: `/var/folders/.../ralph-smoke-20260416-205133`
- **Chat sandbox**: `/private/var/folders/.../ralph-smoke-20260416-205143`

**Both modes**: identical final state — `hello.txt` contains `hello from ralph\nsmoke test passed`, both stories in `tasks.json` have `"passes": true`, loop exited on `ALL_DONE` at iteration 2.

**Change B verified in the raw JSONL** (from `.lnb/entries/cwang31.jsonl`):

- Headless harness entries prefixed `ralph.sh:` (unchanged).
- Chat harness entries prefixed `ralph-lnb:` (confirms the prefix unification reached runtime).
- Chat mode: `ralph-lnb: starting iteration 2/3` timestamp (`20:53:28`) is **after** the iteration-2 prompt was built, confirming the reorder — the harness start entry for iter N is logged *after* `query_recent_history`, so iter N's prompt `recent_history` block excludes its own start. Matches headless.

**Side note (not a blocker)**: `lab-notebook sql` returned stale data against the freshly-written sandboxes (entries with timestamps from a prior run's date, types that don't appear in the JSONL). The JSONL files were correct — the issue is the SQLite index wasn't rebuilt after the new entries appended. `lab-notebook sql` docs say auto-rebuild only triggers when `index.sqlite` is missing, not when stale. Will open a separate issue on the lab-notebook repo to track.

PR #19, opened 2026-06-10
refactor: hand off ralph-lnb prompt via RALPH-PROMPT.md instead of stdout

carbonscott:
## Problem

In the non-headless `/ralph-lnb` skill, the main agent is a pure shuttle but pays for the whole prompt anyway. `ralph-prep.sh` printed the **entire filled prompt** (notebook context + full `tasks.json` + recent history) to stdout, which landed in the main session's transcript as a Bash tool result — and the same text was then passed again as the `Agent(prompt=...)` input. So the full prompt entered the main agent's context **twice per iteration** and accumulated across all iterations (thousands of tokens × N), even though the main agent only needs the subagent's `<promise>` markers.

## Fix (file handoff)

`ralph-prep.sh` now writes the filled prompt to `RALPH-PROMPT.md` in the project dir (parent of `.lnb`) and prints only a one-line status to stdout. The `Agent()` call is a fixed string telling the subagent to read that file. The prompt lands in the subagent's context, where it belongs; the main agent's per-iteration cost drops to ~one status line.

## Changes

- **`cc/ralph-prep.sh`** — redirect `build_prompt` output into `RALPH-PROMPT.md`; emit a single status line; route the notebook `start`-emit confirmation to stderr so stdout stays one line. Header/usage comments updated.
- **`skill/SKILL.md.template`** — Loop step 1 (prompt goes to file, not stdout), Loop step 2 (fixed `Agent()` prompt reading `RALPH-PROMPT.md`), corrected "Context budgeting" section (old ~50-token claim ignored the prompt body), and fixed the inaccurate "headless uses the same ralph-prep.sh" note.
- **`.gitignore`** — ignore `RALPH-PROMPT.md`.

## Scope

`build_prompt` (`shared/ralph-lib.sh`) and the headless runner (`cc-headless/ralph.sh`) are **untouched** — headless calls `build_prompt` directly and depends on its stdout. `ralph-prep.sh` is skill-only.

## Verification

- Stdout is a single status line per iteration (tested two iterations).
- `RALPH-PROMPT.md` is fully filled — 0 leftover `FILL:` markers; iteration 2 picks up prior history.
- `.lnb/` initialized, `start` entry logged.
- `git diff` confirms `shared/ralph-lib.sh` and `cc-headless/` unchanged.
- After `./install.sh`, the regenerated skill reads `RALPH-PROMPT.md` and shows the corrected budgeting section.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .lnb.env
 .lnb/
+RALPH-PROMPT.md
--- a/cc/ralph-prep.sh
+++ b/cc/ralph-prep.sh
@@ -2,8 +2,10 @@
 set -euo pipefail
 
 # Per-iteration bookkeeping + prompt builder for the /ralph-lnb skill.
-# Prints the filled prompt to stdout so a Claude Code session can capture
-# it and pass it to Agent(). Diagnostics go to stderr.
+# Writes the filled prompt to RALPH-PROMPT.md in the project dir and prints
+# only a one-line status to stdout, so the Claude Code session can point a
+# subagent at the file without holding the prompt in its own context.
+# Diagnostics go to stderr.
 
 # --- Defaults ---
 PROMPT_FILE=""
@@ -31,9 +33,10 @@ usage() {
 Usage: ralph-prep.sh [OPTIONS]
 
 Runs one iteration of ralph bookkeeping (archive, notebook init, history
-query, prompt fill) and prints the filled PROMPT.md to stdout. Intended
-to be invoked by a Claude Code session that then passes the stdout to
-the Agent() tool — see the /ralph-lnb skill (skill/SKILL.md.template).
+query, prompt fill) and writes the filled prompt to RALPH-PROMPT.md in the
+project dir, printing only a one-line status to stdout. Intended to be
+invoked by a Claude Code session that then points a subagent at that file
+via the Agent() tool — see the /ralph-lnb skill (skill/SKILL.md.template).
 
 Options:
   --prompt FILE           Custom prompt template (default: repo's shared/PROMPT.md)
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 .lnb.env
 .lnb/
+RALPH-PROMPT.md
```
carbonscott, line 3: Minor: a bare `RALPH-PROMPT.md` pattern matches at any depth, not just the project root where the file is actually written. Fine for an ephemeral artifact — flagging only for intent. Scope to `/RALPH-PROMPT.md` if root-only is meant.

cc/ralph-prep.sh
```diff
@@ -108,15 +111,23 @@ fi
...
 # only prior iterations' entries — matches cc-headless/ralph.sh ordering
 # (see ralph.sh:150-154) so both runners build byte-identical prompts.
 history=$(query_recent_history)
-build_prompt "$history"
 
+# Write the filled prompt to a file in the project dir (parent of the
+# notebook dir) instead of stdout, so the /ralph-lnb main agent never has
+# to hold the full prompt in its context. The subagent reads this file.
+# With the default NOTEBOOK_DIR=.lnb, dirname resolves to "." (project dir).
+PROMPT_OUT="$(dirname "$NOTEBOOK_DIR")/RALPH-PROMPT.md"
```
carbonscott: Minor — path coupling: `dirname "$NOTEBOOK_DIR"` resolves to `.` for the default `.lnb`, matching the subagent's hardcoded `./RALPH-PROMPT.md` (SKILL.md.template, Loop step 2). But `ralph-prep.sh` accepts `--notebook DIR`: with e.g. `--notebook a/b/.lnb`, the prompt is written to `a/b/RALPH-PROMPT.md` while the subagent still reads `./RALPH-PROMPT.md` — a silent mismatch. Latent today since the skill never passes `--notebook` (Loop step 1 omits it). Suggest a one-line comment noting the coupling, or anchor the output to the cwd directly since the subagent path is fixed to cwd anyway.

cc/ralph-prep.sh
```diff
@@ -108,15 +111,23 @@ fi
...
+build_prompt "$history" > "$PROMPT_OUT"
+echo "Wrote iteration $ITERATION prompt to $PROMPT_OUT"
+
+# Route the emit confirmation to stderr so stdout stays a single status line.
```
carbonscott: Nit — comment accuracy: this says "Route the emit confirmation to stderr," but the `>&2` redirects `log_to_notebook`'s start-entry output, not the prep status `echo` on line 126 (which intentionally stays on stdout). Suggest: "Route lab-notebook's start-entry output to stderr so stdout stays a single status line."

cc/ralph-prep.sh
```diff
@@ -108,15 +111,23 @@ fi
...
 
-# --- Emit filled prompt to stdout ---
+# --- Write filled prompt to RALPH-PROMPT.md ---
 # Query history BEFORE logging "start" so the recent_history block sees
 # only prior iterations' entries — matches cc-headless/ralph.sh ordering
 # (see ralph.sh:150-154) so both runners build byte-identical prompts.
 history=$(query_recent_history)
-build_prompt "$history"
 
+# Write the filled prompt to a file in the project dir (parent of the
```
carbonscott: Nit — the "writes to RALPH-PROMPT.md, prints one status line" rationale now appears three times: header (lines 4-8), usage (lines 36-39), and this inline block (lines 120-126). Consistent and helpful, not wrong — just noting the repetition for a short script.

carbonscott (commented):
## Review: hand off ralph-lnb prompt via `RALPH-PROMPT.md`

Correct, minimal refactor that achieves its stated goal: the filled prompt (notebook context + full `tasks.json` + recent history) moves from stdout to `RALPH-PROMPT.md`, so the `/ralph-lnb` main agent no longer holds it in context and the per-iteration cost drops to ~tens of tokens. `set -euo pipefail` + `build_prompt "$history" > "$PROMPT_OUT"` (cc/ralph-prep.sh:125) means a `build_prompt` failure aborts before the status `echo`, so the main agent sees a failed Bash call rather than a silent empty file; the skill's "missing or empty → stop and report" guard (SKILL.md.template, Loop step 2) covers the truncated-file case. The headless runner and `shared/ralph-lib.sh` are correctly left untouched.

**Off-diff follow-up — `README.md:164`:** the Files table still reads `` | `cc/ralph-prep.sh` | … ; stdout is the filled prompt | ``, which now contradicts this PR's behavior. Suggested fix: "…writes the filled prompt to `RALPH-PROMPT.md`; stdout is a one-line status." This is the one item worth fixing before merge — it falls within the PR's documentation surface but isn't in the diff, so it can't be anchored inline.

Inline comments cover the `PROMPT_OUT` path coupling (ralph-prep.sh:124) plus three nits (.gitignore scope, comment wording at ralph-prep.sh:128, repeated rationale).

**Overall: approve with comments.** Only the stale `README.md:164` is worth acting on before merge; everything else is optional polish.

PR #20, opened 2026-06-15
Consolidate ralph state under a gitignored .ralph/ directory

carbonscott:
## What

Consolidates everything ralph generates into one gitignored `.ralph/` directory, closing the state leaks the old layout scattered into the project root. Implements the design in `handoffs/ralph-state-layout.md`.

**Layout after this PR:** `tasks.json` (+ a small `.lnb.env` pointer) stays at the project root; everything ralph generates lives under `.ralph/` — `.ralph/.lnb` (notebook), `.ralph/prompt.md`, `.ralph/state.json` (cursor), `.ralph/archive/`. `.gitignore` shrinks to `.lnb.env` + `.ralph/`. The previously-leaking `.ralph-last-branch` and `archive/` are gone.

## Commits (one per wave, 8 handoffs)

- **Wave A** — byte-identical prompt regression test (`tests/prompt-diff.sh`) + design-spec amendment
- **Wave B** — prompt → `.ralph/prompt.md`; notebook → `.ralph/.lnb` (single `lab-notebook init .ralph`, deletes the brittle mv+sed hack); both runners' `NOTEBOOK_DIR` default flipped together
- **Wave C** — `.ralph-last-branch` → `.ralph/state.json`; resolved archive trigger; stop copying the notebook on archive; record batch completion + refuse all-green re-spin (`--force` to override)
- **Wave D** — per-loop writer `ralph-<branch>` (not `$USER`); opt-in shared notebook via existing `--notebook`; `tests/concurrency.sh`
- **Wave E** — doc-truth sweep (README, skill template, tests/README) to match the landed layout
- **Wave F** — idempotent upgrade shim: migrates old on-disk layout to `.ralph/` once, then no-ops

## Verification

- `tests/prompt-diff.sh`, `tests/archive.sh`, `tests/concurrency.sh`, `tests/upgrade.sh` all exit 0
- `ralph --max-iterations 1` completes end-to-end in a plain non-git dir, producing the target layout with the `ralph-<branch>` writer and zero leak files
- ralph kept git-optional throughout (no `git` in the runtime path)

## ⚠️ Post-merge manual step

`skill/SKILL.md.template` changed, so after merging you must **re-run `install.sh`** (from the canonical clone, against `master`) to re-render `~/.claude/skills/ralph-lnb/SKILL.md`, then restart Claude Code sessions to pick up the new skill.

## Note

The `handoffs/` planning directory (the migration plan + the amended design spec) is intentionally left untracked/out of this PR — it's scaffolding, not product. Say the word if you'd like the design spec committed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,8 +74,8 @@ customize the template. **Upgrading from an older checkout?** A stale
 it if you never customized it, or pass `--prompt ./PROMPT.md`
 explicitly if you did.
 
-The runner auto-initializes `.lnb/` with the coding schema on the first
-iteration — no manual `lab-notebook init` needed.
+The runner auto-initializes `.ralph/.lnb` with the coding schema on the
+first iteration — no manual `lab-notebook init` needed.
 
 Then pick one runner:
 
@@ -94,19 +94,24 @@ Start Claude Code in `acceptEdits` mode, then in the session:
 (Restart any running Claude Code sessions after install — skills load
 at session start.)
 
-`tasks.json` is the entire footprint in your project dir. The prompt
-template, shared lib, helper scripts, and notebook schema all stay in
-the repo and are invoked or sourced by `ralph` / `/ralph-lnb` (or by
-absolute path if you skipped install).
+`tasks.json` is the only file you author and commit in your project
+dir. Everything ralph generates — the notebook, the per-iteration
+prompt, the batch cursor, and archived `tasks.json` snapshots — lives
+under one gitignored `.ralph/` directory, with a small `.lnb.env`
+pointer at the root so a bare `/lnb recall` finds the notebook; a reset
+is just `rm -rf .ralph/`. The prompt template, shared lib, helper
+scripts, and notebook schema all stay in the repo and are invoked or
+sourced by `ralph` / `/ralph-lnb` (or by absolute path if you skipped
+install).
 
 ## How It Works
 
 ```
-tasks.json (what to do)  +  .lnb/ (what happened)  +  PROMPT.md (how to do it)
-           │                       │                           │
-           └───────────┬───────────┘                           │
-                       ▼                                       │
-              runner   builds prompt ◄─────────────────────────┘
+tasks.json (what to do)  +  .ralph/.lnb (what happened)  +  PROMPT.md (how to do it)
+           │                             │                           │
+           └───────────┬─────────────────┘                           │
+                       ▼                                             │
+              runner   builds prompt ◄───────────────────────────────┘
                      │
               ┌──────┴──────────────────────┐
               │  for each iteration:        │
@@ -123,7 +128,7 @@ tasks.json (what to do)  +  .lnb/ (what happened)  +  PROMPT.md (how to do it)
 ## Two runners, same loop
 
 Ralph ships two entry points that share the same `PROMPT.md`, `tasks.json`,
-`.lnb/`, and `archive/` state:
+and `.ralph/` state (notebook, per-iteration prompt, batch cursor, archives):
 
 - **`ralph`** (headless, terminal) — runs in a terminal and spawns a
   fresh `claude -p` per iteration. Truly stateless outer loop. Backed
@@ -161,7 +166,7 @@ The repo splits into four directories so the mode boundary is obvious:
 | File | Purpose |
 |------|---------|
 | `cc-headless/ralph.sh` | Headless runner — uses `claude -p` |
-| `cc/ralph-prep.sh` | Per-iteration bookkeeping + prompt builder invoked by the `/ralph-lnb` skill; writes the filled prompt to `RALPH-PROMPT.md`, stdout is a one-line status |
+| `cc/ralph-prep.sh` | Per-iteration bookkeeping + prompt builder invoked by the `/ralph-lnb` skill; writes the filled prompt to `.ralph/prompt.md`, stdout is a one-line status |
 | `skill/SKILL.md.template` | Skill template — `install.sh` renders this into `~/.claude/skills/ralph-lnb/SKILL.md` with absolute paths |
 | `shared/ralph-lib.sh` | Shared bash helpers sourced by both runners |
 | `shared/PROMPT.md` | Prompt template with `<!-- FILL:xxx -->` markers |
@@ -217,7 +222,7 @@ The `coding-dev.yaml` schema provides entry types tailored for coding:
 
 Query patterns from all iterations:
 ```bash
-LAB_NOTEBOOK_DIR=.lnb lab-notebook sql \
+LAB_NOTEBOOK_DIR=.ralph/.lnb lab-notebook sql \
   "SELECT content FROM entries WHERE type='pattern' ORDER BY ts"
 ```
 
@@ -227,9 +232,11 @@ LAB_NOTEBOOK_DIR=.lnb lab-notebook sql \
 --max-iterations N      Safety cap (default: 10)
 --prompt FILE           Custom prompt template (default: repo's shared/PROMPT.md)
 --task-file FILE        Task file (default: tasks.json)
---notebook DIR          Notebook directory (default: .lnb)
+--notebook DIR          Notebook directory (default: .ralph/.lnb)
 --context SLUG          Notebook context (default: from branch)
---archive-dir DIR       Archive directory (default: archive/)
+--archive-dir DIR       Archive directory (default: .ralph/archive)
+--force                 Re-spin even when this branch's all-green batch
+                        was already recorded complete in .ralph/state.json
 ```
 
 For the Claude Code session runner, invoke it as `/ralph-lnb
@@ -240,8 +247,11 @@ in this repo for the pre-substitution source).
 
 ## Archive
 
-When the `branch` field in `tasks.json` changes between runs, both runners
-archive the previous task file and notebook to `archive/<date>-<branch>/`.
+When the recorded cursor in `.ralph/state.json` shows the `branch` has
+changed since the last run, both runners snapshot the previous
+`tasks.json` — only the task file, not the notebook (it self-archives by
+`context`) — to `.ralph/archive/<date>-<branch>/`. A first run records the
+cursor and archives nothing; a same-branch re-run archives nothing.
 
 ## Shared notebook across loops / worktrees
 
--- a/README.md
+++ b/README.md
@@ -74,6 +74,15 @@ customize the template. **Upgrading from an older checkout?** A stale
 it if you never customized it, or pass `--prompt ./PROMPT.md`
 explicitly if you did.
 
+On the next run, ralph also relocates any old-layout state — `./.lnb`,
+`./.ralph-last-branch`, and `./archive/` — into `.ralph/` and rewrites
+the root `.lnb.env` pointer. The move is one-time and idempotent: once
+the state lives under `.ralph/`, later runs leave it alone. If any of
+those old paths were committed to git, ralph prints a `git rm --cached
+.ralph-last-branch archive` hint — run it to drop those stale index
+entries now that their content lives under the gitignored `.ralph/`
+(ralph never runs git itself).
+
 The runner auto-initializes `.ralph/.lnb` with the coding schema on the
 first iteration — no manual `lab-notebook init` needed.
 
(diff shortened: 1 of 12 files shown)


shared/ralph-lib.sh
```diff
@@ -1,59 +1,245 @@
...
+        return 0
+    fi
+
+    mkdir -p .ralph
+    local migrated_tracked=0
+
+    # 1) Notebook: MOVE ./.lnb -> .ralph/.lnb (never re-init over it). Guarded so
+    #    a second run (target present) is a no-op and a half-migrated tree where
+    #    .ralph/.lnb already exists is left untouched.
+    if [[ -d ./.lnb && ! -e .ralph/.lnb ]]; then
```
carbonscott, line 178: Consider: this step is guarded only on `! -e .ralph/.lnb`. If a prior run already created `.ralph/.lnb` while a stale `./.lnb` still exists, the block is silently skipped and the old `./.lnb` — the user's real history — is orphaned and never queried again. The header comment promises "without data loss" and "converges safely," but this specific both-exist state loses access to history silently. Since this is a one-shot, can't-redo migration, detecting the both-exist case and at minimum warning loudly to stderr would honor that guarantee.

shared/ralph-lib.sh
```diff
@@ -1,59 +1,245 @@
...
+        mv ./.lnb .ralph/.lnb
+        # Rewrite the root pointer so LAB_NOTEBOOK_DIR is the new ABSOLUTE path.
+        # Match the format `lab-notebook init` writes: an `export
+        # LAB_NOTEBOOK_DIR=<abs>` line (lab-notebook parses only that key). We
+        # surgically replace just that line in an existing .lnb.env (preserving
+        # the comment + any LAB_NOTEBOOK_WRITER line); if there's no such line
+        # or no file, (re)write the pointer in the canonical init format.
+        local abs_nb
+        abs_nb="$(cd .ralph/.lnb && pwd)"
+        if [[ -f ./.lnb.env ]] && grep -q '^[[:space:]]*\(export[[:space:]]\+\)\?LAB_NOTEBOOK_DIR=' ./.lnb.env; then
```
carbonscott: Consider: the `.lnb.env` pointer rewrite lives inside the move-guard (`-d ./.lnb && ! -e .ralph/.lnb`). If a run is interrupted after `mv ./.lnb .ralph/.lnb` (line 175) but before this rewrite, the next run sees the guard as false (target exists) and never fixes the pointer — leaving `.lnb.env` aimed at the now-missing `./.lnb`, so a bare `/lnb recall` reads the wrong store. Reconciling the pointer against the actual `.ralph/.lnb` location independent of whether the move happened this invocation would make the shim truly crash-idempotent.

shared/ralph-lib.sh
```diff
@@ -1,59 +1,245 @@
...
     local key="$1"
     jq -r ".$key // empty" "$TASK_FILE" 2>/dev/null || echo ""
 }
 
+# --- Branch slug + writer identity ---
+# Normalize a branch name into a filesystem-safe slug: strip a leading
+# "ralph/" then turn any remaining "/" into "-". E.g. ralph/foo -> foo,
+# feature/x -> feature-x, main -> main. ONE definition shared by both the
+# archive-folder name and the notebook writer id, so they can't drift.
+branch_slug() {
```
carbonscott, line 37: Optional: `branch_slug` isn't injective — `ralph/foo` and `foo` both map to `foo` (likewise `ralph/a/b` and `a-b`). Because the one slug names both the archive folder and the writer id, two genuinely different branches can collide: same-date archive snapshots overwrite, and in shared-notebook mode they share a writer JSONL, defeating the per-loop isolation. Low probability, but the "so they can't drift" comment overstates it — worth documenting the collision or making the slug injective.

shared/ralph-lib.sh
```diff
@@ -1,59 +1,245 @@
...
+        # LAB_NOTEBOOK_DIR=<abs>` line (lab-notebook parses only that key). We
+        # surgically replace just that line in an existing .lnb.env (preserving
+        # the comment + any LAB_NOTEBOOK_WRITER line); if there's no such line
+        # or no file, (re)write the pointer in the canonical init format.
+        local abs_nb
+        abs_nb="$(cd .ralph/.lnb && pwd)"
+        if [[ -f ./.lnb.env ]] && grep -q '^[[:space:]]*\(export[[:space:]]\+\)\?LAB_NOTEBOOK_DIR=' ./.lnb.env; then
+            local tmp_env
+            tmp_env="$(mktemp ./.lnb.env.XXXXXX)"
+            sed "s|^\([[:space:]]*\(export[[:space:]]\+\)\?\)LAB_NOTEBOOK_DIR=.*|\1LAB_NOTEBOOK_DIR=$abs_nb|" \
```
carbonscott: Nit: the `s|...|...|` substitution injects `$abs_nb` unescaped while using `|` as the delimiter. `abs_nb` comes from `cd .ralph/.lnb && pwd`, so a `|` anywhere in a parent directory name (legal, if unusual) would break the expression. A delimiter less likely to appear in a path, or escaping the replacement, would harden this.

tests/concurrency.sh
```diff
@@ -0,0 +1,115 @@
...
+    local branch="$1" message="$2"
+    local BRANCH="$branch"
+    local CONTEXT="$branch"
+    log_to_notebook "impl" "$message"
+}
+
+###############################################################################
+echo "== Two loops on different branches emit into ONE shared notebook =="
+###############################################################################
+emit_as "ralph/loop-a" "loop A: entry from branch a"
```
carbonscott: These two `emit_as` calls run strictly sequentially — there's no `&`/`wait` anywhere in the file — yet the header (lines 4-10) and the test name claim "concurrent writes contention-free." What this actually proves is writer-id *disambiguation* (distinct per-branch JSONL files), which is genuinely useful, but it would still pass if per-writer isolation regressed to a shared-file-with-locking model. Either background the two emits and `wait` to exercise a real race, or reword so the test doesn't over-claim what it verifies.

carbonscott (commented):
## Review Summary

**Overall:** A well-factored, well-tested migration that genuinely improves code health — the state cursor, archive trigger, refuse-respin guard, and per-loop writer identity are clean single-responsibility helpers with good "why" comments, and the regression suite (prompt-diff, archive, upgrade) actually exercises the behavior. Feedback below is substantive but non-blocking; the happy paths (clean old-layout and fresh tree) both work and are covered.

### Key Findings

**Functionality — migration shim robustness (the headline feature)**
The `migrate_old_layout` header promises "without data loss" and "a half-migrated tree converges safely," but two interrupted/edge states don't fully honor that:
- If `.ralph/.lnb` exists *and* a stale `./.lnb` still exists, the notebook step is silently skipped and the old history is orphaned with no warning (ralph-lib.sh:174).
- The `.lnb.env` pointer rewrite lives *inside* the move-guard, so a run interrupted after `mv` but before the rewrite leaves the pointer dangling at the now-missing `./.lnb`, and no later run reconciles it (ralph-lib.sh:184).
Both are narrow (contrived state / sub-ms crash window) and the shim is a one-shot, can't-redo operation — so hardening it (loud warning + idempotent pointer reconciliation) is worth doing in this PR.

**Tests**
- `tests/concurrency.sh` emits the two writers strictly sequentially (no `&`/`wait`), yet the header and name advertise "concurrent writes contention-free." It proves writer-id *disambiguation*, which is valuable, but would still pass under a shared-file-with-locking regression. Either truly background the emits or reword to match what it asserts (concurrency.sh:66).
- `tests/README.md` documents only `prompt-diff.sh`; the three other new scripts (`archive.sh`, `concurrency.sh`, `upgrade.sh`) are undiscoverable from the README (tests/README.md:12).
- Coverage gaps worth a line each (non-blocking): partial-green *non*-refusal (a half-done batch must not be refused even if `completed_branch` matches), and `completed_branch` preservation across an ordinary same-branch re-run.

**Documentation — completeness**
The automatic old-layout migration is the headline behavioral change, yet the only "Upgrading from an older checkout?" note covers solely the stale `./PROMPT.md`. Existing users get no prose telling them `./.lnb`, `./.ralph-last-branch`, and `./archive/` are auto-relocated, nor that they may need `git rm --cached` for old paths that leaked into git (README.md:75).

**Minor**
- `branch_slug` is non-injective: `ralph/foo` and `foo` collide to the same slug, which names both the archive folder and the writer id — the comment "so they can't drift" slightly overstates the guarantee (ralph-lib.sh:34).
- The `.lnb.env` pointer rewrite uses `|` as the `sed` delimiter while substituting an absolute path; a `|` in a parent directory name would break it (ralph-lib.sh:187).
- `rm -rf .ralph/` as the documented "reset" leaves the root `.lnb.env` dangling until the next run (README.md:102).
- `skill/SKILL.md.template` frontmatter `argument-hint` (line 5, outside the diff hunk) still omits the new `force` flag that the same PR adds to the args table, examples, and parser — so `/ralph-lnb force` won't surface in completion.

### What's done well
- The byte-identical `prompt-diff.sh` reconstructs both runners' prompts and compares with `cmp -s` (no normalization) — a genuine, non-tautological regression guard.
- `upgrade.sh` idempotency coverage is the strongest part of the suite: second-run no-op, half-migrated convergence, and an inode check.
- The refuse-respin guard runs *before* archive/ensure_notebook and exits 3 without mutating state — a clean signal for the loop to stop, with `--force` to override.
- `.gitignore` collapses to the single `.ralph/` umbrella correctly, with `.lnb.env` retained.
- The doc-truth path sweep across README and SKILL.md.template is thorough — no stale `.lnb/`, `RALPH-PROMPT.md`, `.ralph-last-branch`, or `archive/` references survive.

**Event: COMMENT** — improves code health; the items above are worth addressing but none block merge. 9 inline comments below.

tests/README.md
```diff
@@ -9,6 +9,12 @@ Exercises both modes so you can verify the `/ralph-lnb` skill and
 
 - `tasks.json` — 2-story fixture (create `hello.txt`, append a line)
 - `setup-sandbox.sh` — scaffolds a timestamped sandbox under `$TMPDIR`
+- `prompt-diff.sh` — exit-code regression test asserting the two runners
```
carbonscott, line 12: This PR adds four test scripts but the README's file list documents only `prompt-diff.sh`. `archive.sh`, `concurrency.sh`, and `upgrade.sh` are left undiscoverable — a maintainer reading this has no way to know they exist or how to run them. Add entries for the other three (and a one-line "run all" note) so the README documents the full suite it ships.

README.md
```diff
@@ -74,8 +74,8 @@ customize the template. **Upgrading from an older checkout?** A stale
 it if you never customized it, or pass `--prompt ./PROMPT.md`
 explicitly if you did.
```
carbonscott, line 75: Completeness: this is the only "Upgrading from an older checkout?" guidance, and it covers solely the stale `./PROMPT.md`. But the headline change of this PR is an automatic `migrate_old_layout` shim that relocates `./.lnb`, `./.ralph-last-branch`, and `./archive/` into `.ralph/` and even prints `git rm --cached` guidance to stderr. Existing users — the exact audience this paragraph addresses — get no prose about that migration. Consider extending this note to mention the auto-relocation and the possible `git rm --cached` cleanup for old paths that leaked into git.

README.md
```diff
@@ -94,19 +94,24 @@ Start Claude Code in `acceptEdits` mode, then in the session:
...
-`tasks.json` is the entire footprint in your project dir. The prompt
-template, shared lib, helper scripts, and notebook schema all stay in
-the repo and are invoked or sourced by `ralph` / `/ralph-lnb` (or by
-absolute path if you skipped install).
+`tasks.json` is the only file you author and commit in your project
+dir. Everything ralph generates — the notebook, the per-iteration
+prompt, the batch cursor, and archived `tasks.json` snapshots — lives
+under one gitignored `.ralph/` directory, with a small `.lnb.env`
+pointer at the root so a bare `/lnb recall` finds the notebook; a reset
+is just `rm -rf .ralph/`. The prompt template, shared lib, helper
```
carbonscott: Optional: "a reset is just `rm -rf .ralph/`" leaves the root `.lnb.env` pointer (established earlier in this same sentence as living *outside* `.ralph/`) dangling at the now-deleted `.ralph/.lnb`. Re-running ralph recovers it, but a bare `/lnb recall` in the window before the next run hits a stale pointer — the very thing the pointer exists to prevent. Either note `rm -rf .ralph/ .lnb.env` for a full reset, or state that the pointer is regenerated on the next run.

tests/prompt-diff.sh
```diff
@@ -0,0 +1,188 @@
...
+            else
+                CONTEXT="ralph-dev"
+            fi
+        fi
+
+        # Mirror ralph.sh's main path: bookkeeping, then RECALL + build.
+        archive_previous_run
+        ensure_notebook
+        local history
+        history=$(query_recent_history)   # empty notebook -> constant string
```
carbonscott, line 100: FYI: nicely done — this reconstructs path A by sourcing the lib and calling `build_prompt` exactly as `ralph.sh` does, runs the real `ralph-prep.sh` for path B, and compares with `cmp -s` (no normalization). Neutralizing the only time-varying input with fresh empty notebooks, and deliberately not hardcoding the empty-history string, keeps it correct as that string evolves. This would genuinely fail if the two runners' prompts diverged.

PR #21, opened 2026-07-07
Migrate harness to the minimal lnb CLI (note/log/retract)

carbonscott:
## What

Migrates the ralph-wiggum-lnb harness off the old multi-command `lab-notebook` CLI (`emit`/`sql`/`init`/`schema`) onto the new **minimal single-file `lnb`** CLI (`note`/`log`/`retract`). The min CLI now lives on `main` of `carbonscott/lab-notebook` (jq is the read path — no SQL, no schema, no config).

## Mapping

| Old | New |
|-----|-----|
| `lab-notebook emit --context C --type T --branch B --tags X "msg"` | `lnb note "msg" --type T --context C branch=B tags=X` |
| `lab-notebook sql "SELECT … FROM entries WHERE …"` | `lnb log \| jq '<filter>'` |
| `lab-notebook init … --template-path coding-dev.yaml` | *(dropped — auto-creates on first `lnb note`)* |
| `LAB_NOTEBOOK_DIR` / `LAB_NOTEBOOK_WRITER` | `LNB_DIR` / `LNB_WRITER` |
| `.lnb.env` pointer file | *(removed — harness passes `LNB_DIR` per call)* |

## Changes

- **`shared/ralph-lib.sh`** — `log_to_notebook` → `lnb note`; `query_recent_history` → `lnb log | jq -c --arg ctx … | tail -n 10 | tac` (context filter injection-safe via `--arg`, 200-char trim, DESC order; `(no history yet)` fallback restored because `lnb log` is silent on an empty notebook). `ensure_notebook` drops `init` for `mkdir -p`. `write_notebook_pointer` + all `.lnb.env` logic removed; `migrate_old_layout` keeps the `./.lnb → .ralph/.lnb` move minus the pointer reconcile.
- **`shared/PROMPT.md`**, **`skill/SKILL.md.template`** — all `emit`/`sql` examples → `note`/`jq` (interrupted-work `NOT IN` → jq slurp; `LIKE`/FTS → `test()`). The template renders to `~/.claude/skills/ralph-lnb/SKILL.md`, so it must be clean.
- **`shared/coding-dev.yaml`** — deleted (only fed `lab-notebook init --template-path`).
- **`README.md`**, **`tests/README.md`** — install/prereqs + examples updated; schema section reframed as conventional `+type` values (lnb validates nothing by design).
- **`tests/{concurrency,upgrade}.sh`** — reworked off SQL/`.lnb.env` onto `lnb log | jq` and the top-level `.lnb/<writer>.jsonl` layout. `upgrade.sh` drops assertions about machinery that no longer exists.

## Verification

- `grep -rnE 'lab-notebook |LAB_NOTEBOOK' shared/ cc/ cc-headless/ skill/ tests/` → **no matches**.
- All four harness tests pass against real `lnb`: **concurrency**, **upgrade**, **prompt-diff**, **archive**.


shared/ralph-lib.sh
```diff
@@ -237,74 +217,55 @@ migrate_old_layout() {
...
+    if command -v lnb &>/dev/null && [[ -d "$NOTEBOOK_DIR" ]]; then
+        # `lnb log` streams every record as JSONL ascending by ts. Filter to this
+        # context, trim content to 200 chars, keep the 10 most recent (tail), then
+        # flip to descending (tac) to match the old ORDER BY ts DESC LIMIT 10.
+        # --arg keeps CONTEXT out of the jq program (no injection/quoting hazard).
+        local history
+        history=$(LNB_DIR="$NOTEBOOK_DIR" lnb log 2>/dev/null \
+            | jq -c --arg ctx "$CONTEXT" \
+                'select(.context==$ctx) | {ts,type,issue,content:(.content[0:200])}' \
+            | tail -n 10 | tac) || true
```
carbonscott: Optional (portability): `tac` is GNU coreutils and is absent on macOS/BSD by default (and there's no `tail -r` fallback here). Under `set -euo pipefail`, a missing `tac` makes this stage exit 127; pipefail fails the pipeline, `|| true` swallows it, and `history` ends up empty, so on those platforms `query_recent_history` silently degrades to always `(no history yet)` and leaks a `tac: command not found` to the user's terminal every iteration (only `lnb log`'s stderr is redirected, not this stage's). Consider doing the reverse in jq and dropping both `tail` and `tac`, e.g. slurp with `-s` and emit `... | .[-10:] | reverse | .[]`.

(For the record: SIGPIPE is *not* a concern here — `tail -n 10` on a pipe reads to EOF before emitting, so it never closes the pipe early on `lnb`/`jq`.)

shared/ralph-lib.sh
```diff
@@ -237,74 +217,55 @@ migrate_old_layout() {
...
-        LAB_NOTEBOOK_DIR="$NOTEBOOK_DIR" LAB_NOTEBOOK_WRITER="$(notebook_writer)" \
-            lab-notebook emit \
-            --context "$CONTEXT" --type "$entry_type" \
-            --branch "${BRANCH:-}" --tags "ralph-harness" \
-            "$message" 2>/dev/null || true
+        # lnb subprocess. This keeps several worktrees/loops pointed at one
+        # shared notebook contention-free (per-writer JSONL). Content string
+        # LEADS; branch/tags are key=value writer fields, not flags.
+        LNB_DIR="$NOTEBOOK_DIR" LNB_WRITER="$(notebook_writer)" \
+            lnb note "$message" \
```
carbonscott, line 248: Nit (latent — not triggered by current callers): lnb's grammar treats a leading `+`/`#`/`@` as a type/context sigil and a leading `key=value` token as a field, and it has no `--` end-of-options separator. Since `$message` is the only positional here, any `$message` that begins with `+ # @ -` or matches `^[A-Za-z_][\w.-]*=` is consumed as a sigil/field, leaving zero positionals → `lnb note` dies `nothing to log` (exit 1), which `2>/dev/null || true` then swallows — the entry silently vanishes with no diagnostic (verified against `lnb.py`'s `parse()`/`cmd_note`). Every current caller passes a fixed `ralph.sh:`/`ralph-lnb:` string, so this is safe today. Worth a one-line note on the function that `$message` must not lead with a sigil, precisely because the swallow makes any future violation invisible.

shared/ralph-lib.sh
```diff
@@ -164,23 +163,19 @@ archive_previous_run() {
 migrate_old_layout() {
     # Nothing to do unless at least one OLD-layout artifact is present. Avoids
     # creating a stray .ralph/ on fresh projects (the shim must be inert there).
-    if [[ ! -e ./.lnb && ! -e ./.lnb.env && ! -e ./.ralph-last-branch && ! -e ./archive ]]; then
+    if [[ ! -e ./.lnb && ! -e ./.ralph-last-branch && ! -e ./archive ]]; then
```
carbonscott, line 166: Nit: dropping `./.lnb.env` from this trigger guard is correct (lnb no longer uses it), but a side effect is that a stale root `./.lnb.env` from an old install is now never swept — if it's the *only* remaining old artifact, the shim early-returns and leaves it forever. It's inert (min-lnb never reads it, and the harness always passes `LNB_DIR`), so this is harmless; but an opportunistic `rm -f ./.lnb.env` inside the migration would leave a genuinely clean tree. Optional.

tests/concurrency.sh
```diff
@@ -50,20 +51,27 @@ cd "$SHARED_ROOT"
...
+# derived by notebook_writer (BRANCH -> ralph-<slug>, the runners' real path)
+# and set as LNB_WRITER on the `lnb note` call, so each loop appends to its own
+# per-writer JSONL under the shared notebook.
 emit_as() {
     local branch="$1" message="$2"
     local BRANCH="$branch"
     local CONTEXT="$branch"
-    log_to_notebook "impl" "$message"
+    LNB_DIR="$NOTEBOOK_DIR" LNB_WRITER="$(notebook_writer)" \
+        lnb note "$message" --type impl --context "$CONTEXT" \
```
carbonscott, line 73: FYI: `emit_as` now inlines the `lnb note` command instead of calling `log_to_notebook`, so no test exercises the real helper's arg ordering or its `2>/dev/null || true`. The rationale is sound (you want a failed emit to surface through `wait`, and `log_to_notebook` deliberately swallows failures), but the inlined command line can now drift from the real one undetected. Consider a short comment noting the intentional divergence, or a separate assertion that `log_to_notebook` itself writes a well-formed record.

carbonscott (commented):
## Review Summary

**Overall:** The migration is functionally sound and the bash is careful. `set -euo pipefail` is respected throughout, the `history=$(...) || true` capture is correct (and `local history` is declared on its own line so `local` can't mask the exit status), the jq `--arg` keeps `CONTEXT` out of the program (injection-safe), and `migrate_old_layout` stays idempotent with the both-exist warning intact. The reworked tests use the `grep && fail` / `... || fail` set -e idioms correctly and the inode-based idempotence checks are a nice touch. No blockers; all findings are minor or latent.

### Key findings (shell correctness & operational safety)
- **Portability — `tac`:** new hard dependency on GNU `tac` in `query_recent_history` (`shared/ralph-lib.sh:261`). Silently degrades to `(no history yet)` plus stderr noise on macOS/BSD. Inline comment; suggest reversing in jq.
- **Latent misparse — `log_to_notebook`:** a `$message` leading with `+ # @ -` or matching `key=value` makes `lnb note` exit 1, which `2>/dev/null || true` swallows, so the entry silently vanishes. Current callers all pass fixed strings, so it's latent. Inline on `shared/ralph-lib.sh:244`.
- **Cleanup — leftover `.lnb.env`:** the migration no longer sweeps a stale root `.lnb.env` (inert, harmless). Inline on `shared/ralph-lib.sh:166`.
- **Test coverage:** `concurrency.sh` now bypasses `log_to_notebook`; drift risk. Inline on `tests/concurrency.sh:73`.

### Not in the diff (noted here, not inline)
- **README still documents the removed `.lnb.env` pointer.** This PR's own mapping table lists `.lnb.env` as removed, yet `README.md:75-77` still says ralph *"rewrites the root `.lnb.env` pointer"* and `README.md:107-110` describes *"a small `.lnb.env` pointer at the root"* with a reset of `rm -rf .ralph/ .lnb.env`. min-lnb never creates `.lnb.env` (discovery is `$LNB_DIR`, which the harness passes on every call), so both passages now describe machinery that no longer exists. Recommend updating them.
- **Stale `lab-notebook` comments in the (untouched) runners:** `cc-headless/ralph.sh:32` (*"a lab-notebook"*) and `cc/ralph-prep.sh:171` (*"Route lab-notebook's start-entry output to stderr"*). The PR's "no matches" check used the pattern `lab-notebook ` with a trailing space, which misses `lab-notebook's` / `lab-notebook.`. Cosmetic, but worth a sweep.

### Verified OK (no action)
- `set -euo pipefail` + `history=$(...) || true`: correct — a pipeline failure can't abort the run, and `tail -n 10` reads to EOF so there is no early-close SIGPIPE hitting `lnb`/`jq`.
- Empty / not-available fallback: correct for both the empty-notebook branch (`lnb log` prints nothing on stdout, so `history` is empty and the `(no history yet)` fallback fires) and the `else` (`notebook not available`) branch.
- `ensure_notebook` ordering: both runners call `ensure_notebook` before any log/query, and `lnb note` itself creates the dir via `find_notebook(create=True)`, so the `[[ -d ]]` guards can't be reached with a missing dir.
- `migrate_old_layout`: both-exist warning, partial-migration convergence, and idempotence all still hold after removing the pointer reconcile.

_4 inline comments left on specific lines._

shared/ralph-lib.sh
```diff
@@ -237,74 +217,55 @@ migrate_old_layout() {
...
-            2>/dev/null || echo "(no history yet)"
+    if command -v lnb &>/dev/null && [[ -d "$NOTEBOOK_DIR" ]]; then
+        # `lnb log` streams every record as JSONL ascending by ts. Filter to this
+        # context, trim content to 200 chars, keep the 10 most recent (tail), then
+        # flip to descending (tac) to match the old ORDER BY ts DESC LIMIT 10.
+        # --arg keeps CONTEXT out of the jq program (no injection/quoting hazard).
+        local history
+        history=$(LNB_DIR="$NOTEBOOK_DIR" lnb log 2>/dev/null \
+            | jq -c --arg ctx "$CONTEXT" \
+                'select(.context==$ctx) | {ts,type,issue,content:(.content[0:200])}' \
```
carbonscott: FYI — verified this reproduces the old `ORDER BY ts DESC LIMIT 10` faithfully: filtering happens before `tail -n 10`, so it's the 10 most-recent *matching* rows, and `.content[0:200]` matches `substr(content,1,200)` on codepoints (checked multibyte). One behavior note: this view is fail-open — a *retracted* entry whose context matches still shows up here (the tombstone carries no `context`, so it's correctly excluded, but the entry it retracts is not). That's consistent with lnb's documented model and fine for a recent-history digest; just flagging it in case parity with a live view was intended.

shared/PROMPT.md
```diff
@@ -59,32 +58,32 @@ notebook has the *reasoning, failures, and patterns* behind them.
...
 
--- Learn patterns discovered by prior agents
-SELECT content FROM entries WHERE context='<!-- FILL:context -->' AND type='pattern' ORDER BY ts
+# Learn patterns discovered by prior agents (log is already ascending by ts)
+lnb log | jq -r --arg c '<!-- FILL:context -->' 'select(.context==$c and .type=="pattern").content'
 
--- Find interrupted work to resume
-SELECT e.issue, e.content FROM entries e WHERE context='<!-- FILL:context -->' AND type='start' AND issue NOT IN (SELECT issue FROM entries WHERE context='<!-- FILL:context -->' AND type='done')
+# Find interrupted work to resume (a start with no matching done)
+lnb log | jq -s --arg c '<!-- FILL:context -->' '[.[]|select(.context==$c and .type=="done").issue] as $d | .[] | select(.context==$c and .type=="start" and (.issue|IN($d[])|not)) | {issue,content}'
```
carbonscott, line 75: FYI — `(.issue|IN($d[])|not)` is a correct `NOT IN`, and it's actually more robust than the SQL it replaces: in SQLite, `issue NOT IN (SELECT issue …)` returns *zero* rows if any `done.issue` is NULL. The jq form has no such trap. Nice.

shared/PROMPT.md
```diff
@@ -59,32 +58,32 @@ notebook has the *reasoning, failures, and patterns* behind them.
...
 
--- Check what files were changed for a related story
-SELECT ts, type, content FROM entries WHERE context='<!-- FILL:context -->' AND files_changed LIKE '%auth.py%'
+# Check what files were changed for a related story
+lnb log | jq -c --arg c '<!-- FILL:context -->' 'select(.context==$c and ((.files_changed//"")|test("auth.py"))) | {ts,type,content}'
```
carbonscott: Nit: `test("auth.py")` treats the argument as a **regex**, so the `.` matches any character (e.g. `authXpy` would match) — the old `LIKE '%auth.py%'` treated `.` literally. Harmless for this example, but if you want literal substring parity use `(.files_changed//"") | index("auth.py")` or escape the dot: `test("auth\\.py")`. Good call guarding the missing key with `//""` so a record without `files_changed` doesn't error.

skill/SKILL.md.template
```diff
@@ -96,21 +96,21 @@ For `i` in `1..max-iterations`:
...
+     - Call `Bash(...)` to log a blocker entry (same shape as above but `+blocker` and `'ralph-lnb: iteration i ended without promise'`).
      - Break the loop. Go to "Post-loop".
 
 ### Post-loop
 
 1. If the loop ended on `ALL_DONE`, optionally query the notebook for a summary, using the `<context>` derived in Setup step 2:
 
    ```
-   Bash("LAB_NOTEBOOK_DIR=<notebook> lab-notebook sql \"SELECT ts, type, issue, substr(content,1,200) FROM entries WHERE context='<context>' AND type IN ('start','done','impl','blocker') ORDER BY ts\"")
+   Bash("LNB_DIR=<notebook> lnb log | jq -c --arg c '<context>' 'select(.context==$c and (.type|IN(\"start\",\"done\",\"impl\",\"blocker\"))) | {ts,type,issue,content:(.content[0:200])}'")
```
carbonscott, line 113: Nit: the `\"start\"` literals sit inside a shell **single-quoted** jq program, so they're only correct once the illustrative `Bash("…")` wrapper is un-escaped (`\"`→`"`). Decoded, it runs clean (verified). But if an agent passes the `\"` through verbatim, they land inside single quotes and jq fails to compile (`syntax error, unexpected INVALID_CHARACTER`). Same nested-quote convention as the old SQL example, so not a regression — just consider noting that the outer quotes are illustrative.

tests/README.md
```diff
@@ -24,7 +24,7 @@ Exercises both modes so you can verify the `/ralph-lnb` skill and
   branches pointed at one shared notebook each emit, concurrently, under their
   own `ralph-<branch>` writer, landing in distinct `entries/ralph-*.jsonl`
```
carbonscott: Nit: stale path — the min CLI writes per-writer files at the top level (`.lnb/ralph-*.jsonl`), not under an `entries/` subdir. `concurrency.sh` itself was correctly updated to assert the top-level layout; this doc line should match.

carbonscott (commented):
## Review Summary

**Overall:** The jq translations are faithful to the old SQL semantics — I verified each one against a scratch notebook (including multibyte content, records missing `issue`, and a `_retract` tombstone) and re-ran the reworked `concurrency.sh` and `upgrade.sh`; both pass. The one real problem is doc-only: **README.md still documents the `.lnb.env` pointer that this very PR deletes**, which contradicts the code and will mislead users about discovery and reset.

### Functionality / jq fidelity (verified good)

- **`query_recent_history`** (`shared/ralph-lib.sh`): `tail -n 10 | tac` over the ascending `lnb log`, filtered *before* the tail, correctly reproduces `WHERE context=? ORDER BY ts DESC LIMIT 10` (10 most-recent-matching, descending). `.content[0:200]` slices by codepoints and matches `substr(content,1,200)` for multibyte/short strings (checked a 226-codepoint café/touché string → 200). `.issue` on records lacking the key yields `null` (no throw). Tombstones are excluded for free because they carry no `context`.
- **Interrupted-work slurp** (`PROMPT.md`): `(.issue|IN($d[])|not)` is a correct `NOT IN`, and as a bonus it sidesteps SQL's NULL-in-`NOT IN` trap (a single NULL `done.issue` would have made the old query return zero rows).
- **`test()` translations**: `files_changed`/`tags` are treated as the comma-joined **strings** they now are (no array ops) — correct. See the nit on regex-vs-LIKE below.
- **Post-loop summary** (`SKILL.md.template`): `(.type|IN("start","done","impl","blocker"))` is correct; the decoded command runs clean. See the escaping nit.

### Documentation (blocking-ish, doc-only)

**`README.md` still describes the removed `.lnb.env` pointer** — these lines are outside the diff hunks so I could not inline them:

> **`README.md` line 76-77** — "…into `.ralph/` and **rewrites the root `.lnb.env` pointer**." This PR removes `write_notebook_pointer` and the pointer-reconcile block from `migrate_old_layout`; the shim no longer creates or rewrites `.lnb.env`. Please drop this clause.

> **`README.md` line 107-111** — "…with a small **`.lnb.env` pointer at the root so a bare `/lnb recall` finds the notebook**; a full reset is **`rm -rf .ralph/ .lnb.env`** … the next run **regenerates it**." None of this holds now: no `.lnb.env` is ever written, nothing regenerates it, and the minimal `lnb` has no `recall` subcommand (only `note`/`log`/`retract`). Discovery is now `LNB_DIR`, passed explicitly on every harness call. Suggest: reset is just `rm -rf .ralph/`; drop the pointer and `/lnb recall` entirely.

> **`README.md` line 111-112** — "…the prompt template, shared lib, helper scripts, and **notebook schema** all stay in the repo." `shared/coding-dev.yaml` is deleted in this PR, so "notebook schema" no longer applies — drop it.

**`tests/concurrency.sh` line 83** (outside the diff hunks) — the comment still reads "distinct `entries/ralph-<branch>.jsonl`", but the min CLI stores per-writer files at the top level `.lnb/<writer>.jsonl` (the code below was correctly updated). Please fix the stale comment.

### Tests

Good rework. `upgrade.sh`'s deletion of the `.lnb.env`/`init`/PART-B machinery is justified (that machinery is gone), and switching the idempotence/partial-convergence checks to inode comparison is a clean way to assert "not re-moved." `concurrency.sh` now asserts the real top-level `.lnb/<writer>.jsonl` layout and slurps `lnb log` with jq. Both pass here against real `lnb`.

5 inline comments left (1 doc nit, 3 FYI/nit on jq examples, 1 escaping nit).

carbonscott:
## Review disposition (2 independent subagent reviews)

Two independent reviews were posted — **shell-correctness** (review 4640702769) and **jq-fidelity** (review 4640707498). No blockers or majors from either. Every finding dispositioned below; fixes landed in 68de53a.

| # | Finding | Severity | Disposition |
|---|---------|----------|-------------|
| F1 | `query_recent_history` used `tail\|tac` (GNU-only; stderr leak under pipefail) | minor | **Fixed** — reverse inside jq, no coreutils dep |
| F2 | `log_to_notebook` content-leads misparse is latent + silently swallowed | minor | **Fixed** — documented the constraint on the helper |
| F3 / R2-1 | README still documents removed `.lnb.env` pointer / `/lnb recall` / notebook schema | minor→major | **Fixed** — rewrote both passages to `LNB_DIR`-per-call; reset is `rm -rf .ralph/` |
| F4 | Stale `lab-notebook` comments in `cc-headless/ralph.sh`, `cc/ralph-prep.sh` | nit | **Fixed** — comments updated |
| F5 | Migration doesn't sweep a lone leftover `.lnb.env` | nit | **Declined** — inert (lnb never reads it); re-adding it to the migrate trigger reintroduces the `.lnb.env` coupling the migration deliberately sheds |
| F6 | `concurrency.sh` inlines the write instead of calling `log_to_notebook` | nit | **Declined** — deliberate: the test wants a failed write to surface via `wait`, not be swallowed by `log_to_notebook`'s `2>/dev/null \|\| true` |
| R2-2 | `concurrency.sh:83` comment says `entries/ralph-<branch>.jsonl` | minor | **Fixed** — top-level `.lnb/ralph-<branch>.jsonl` |
| R2-3 | `tests/README.md:25` stale `entries/ralph-*.jsonl` | minor | **Fixed** — `.lnb/ralph-*.jsonl` |
| R2-4 | `PROMPT.md` `test("auth.py")` regex dot matches any char | nit | **Fixed** — `test("auth\\.py")` |
| R2-5 | `SKILL.md.template` illustrative `Bash("...")` quote depth | nit | **Declined** — decoded form runs; same convention as the prior SQL example, not a regression |
| R2-6 | Reads are fail-open on retracted entries | info | **Declined** — consistent with lnb's documented model; ralph never retracts, so a digest never shows a tombstoned entry in practice |

All four harness tests (concurrency, upgrade, prompt-diff, archive) pass against real `lnb` after the fixes; the migration scope grep is clean.

